PR #57, opened 2018-05-22
Tracking boundaries during interactive proof refinement

jonsterling:
This is not even close to ready, but I'm opening it so that I can see all my gnarly changes in one place. What a pain in the neck! I know how to write these proof state transitions on paper in a way that is nice, but I don't yet know how to program them cleanly.


jonsterling:
done, thanks!

PR #90, opened 2018-06-11
coe of V

### LINKED ISSUE #42: coe in V

favonia:
Fix RedPRL/redtt#42. The current code is actually incorrect---the 0/1 cases only make sense when `x=y` in `x.V_y(...)`.

What is the immortal way to have alternative dynamics? @jonsterling


jonsterling (changes requested):

jonsterling (approved):
Your code is looking nice! I'm obviously very happy to be trusting you about the details of the coe algorithm(s) implemented here.

PR #138, opened 2018-07-02
eta law for v types

jonsterling:
This PR includes the eta law for V-types, but it also includes some a more immortal version of the elaborator that lets the user control intermediate states where boundaries don't match *yet*. But it also uncovers some bugs in other parts of the code.

### DIFF
--- a/connection.red
+++ b/connection.red
@@ -18,7 +18,7 @@ let connection/or
    end
  =
  λ i j →
-  ; this is an example of something that is much nicer here than in redprl and yacctt.
+  ; this is an example of something that is much nicer here than in redprl.
   ; we can define using line types all the faces of the composition at once.
   ; definitional equivalence kicks in to make this work.
   let face : Line (Line A) =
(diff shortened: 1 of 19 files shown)


jonsterling:
In addition to adding eta for V-types, this branch fixes some bugs that have to do with the behavior of globals & restrictions. Hopefully it will all become cleaner later when we adopt Favonia's Thought.

This branch also includes a less frustrating refinement/elaboration experience. Like RedPRL, the redtt elaborator used to suspend typechecking problems which failed because of mismatched boundaries, generating unification subgoals in the hope that they might be solved later.

This made it really hard to tell when you had an incorrect proof vs a typo. What redtt does now is actually requires that these unification subgoals should be dischargable immediately, and it will fail if they can't be solved right away. Now, during the proof refinement process you may be working on something where the boundaries don't yet match, and that's ok! But if you are doing that, you should surround the part of your proof with the incorrect boundary with `?{ ... }`; this has the effect of allowing you to at least make an element of the right type, and then once you have got the boundaries right, you can remove the braces.

The braces have the effect of hiding your intermediate node from definitional equivalence, which is crucial for maintaining the invariant that all intermediate proof states are well-typed.

PR #155, opened 2018-07-11
Fix bugs

jonsterling:
This is very mortal right now

### DIFF
--- a/nat.red
+++ b/nat.red
@@ -43,5 +43,5 @@ let plus/suc/r (n : nat) : (m : nat) -> Path nat (plus n (suc m)) (suc (plus n m
 let plus/comm (m : nat) : (n : nat) -> Path nat (plus n m) (plus m n) =
   nat-rec m with
   | zero => plus/unit/r
-  | suc (m => plus/comm/m) => lam n -> trans nat (suc (plus m n)) (plus/suc/r n m) (lam i -> suc (plus/comm/m n i))
+  | suc (m => plus/comm/m) => lam n -> trans _ (suc (plus m n)) (plus/suc/r n m) (lam i -> suc (plus/comm/m n i))
   end
--- a/nat.red
+++ b/nat.red
@@ -3,13 +3,11 @@ import path
 let pred (x : nat) : nat =
   nat-rec x with
   | zero ⇒ zero
-  | suc (n ⇒ ih) ⇒ n
+  | suc n ⇒ n
   end
 
-; broken
-let nat/refl (x : nat) : Path nat x x =
+let pred/succ/zero (x : nat) : Path nat x (pred (suc x)) =
   nat-rec x with
-  | zero ⇒ λ _ → zero
+  | zero ⇒ λ i → zero
   | suc n ⇒ λ _ → suc n
   end
-
--- a/nat.red
+++ b/nat.red
@@ -1,0 +1,15 @@
+import path
+
+let pred (x : nat) : nat =
+  nat-rec x with
+  | zero ⇒ zero
+  | suc (n ⇒ ih) ⇒ n
+  end
+
+; broken
+let nat/refl (x : nat) : Path nat x x =
+  nat-rec x with
+  | zero ⇒ λ _ → zero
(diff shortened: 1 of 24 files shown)


src/core/Tm.ml
```diff
@@ -617,22 +617,22 @@ let rec bindn xs txs =
...
   let rec go k xs tyxs sysxs =
     match xs with
-    | Emp -> tyxs, sysxs
-    | Snoc (xs, x) ->
+    | [] -> tyxs, sysxs
+    | x :: xs ->
       go (k + 1) xs (close_var x (n - k) tyxs) (map_tm_sys (close_var x (n - k)) sysxs)
   in
-  NB (List.map Name.name xs_l, go 0 xs tyxs sysxs)
+  NB (List.map Name.name xs_l, go 0 xs_l tyxs sysxs)
```
jonsterling: I don't understand why this commit is correct, nor, if it is, why it doesn't affect any of our examples except for the nat-rec one.

jonsterling: In fact, I think the change is incorrect, but that something somewhere is written backwards.

jonsterling:
Basically the problem is that the code is hopelessly inconsistent about which direction multibinders are going, and I'm not mentally equipped to figure it out right now...

jonsterling:
I think we're less owned now.

nat.red
```diff
@@ -0,0 +1,13 @@
+import path
+
+let pred (x : nat) : nat =
+  nat-rec x with
+  | zero ⇒ zero
+  | suc n ⇒ n
+  end
+
+let pred/succ/zero (x : nat) : Path nat x (pred (suc x)) =
```
jonsterling: By the way, it's worth observing that this holds by *case*, without needing induction. This is because in each case, the equation holds definitionally (but not prior to the case split).

src/core/Quote.ml
```diff
@@ -324,6 +325,21 @@ struct
...
+      let env' = Env.succ env in
+      let mot = equate_ty env' vmot0 vmot1 in
+      let vmot_ze = inst_clo rec0.mot @@ make Zero in
+      let zcase = equate env vmot_ze rec0.zcase rec1.zcase in
+      let ih = generic env' @@ vmot0 in
+      let vmot_su = inst_clo rec0.mot @@ make @@ Suc var in
+      let scase0 = inst_nclo rec0.scase [var; ih] in
+      let scase1 = inst_nclo rec1.scase [var; ih] in
+      let scase = equate (Env.succ env') vmot_su scase0 scase1 in
+      let frame = Tm.NatRec {mot = Tm.B (clo_name rec0.mot, mot); zcase; scase = Tm.NB (Emp #< None #< None, scase)} in
```
favonia: Why can't we use `nclo_names` here?

jonsterling: I didn't know it existed and I was being lazy;)  let's do it.

favonia:
Sorry I forgot to commit the code which quotes `nat-rec` back to the Realm of Syntax.

src/core/Quote.ml
favonia: It did not exist---I needed to add it. It is simple though.

favonia: My question was a suggestion concisely summarized as a code snippet.

jonsterling: Ah!

jonsterling:
I am about to lift off in the airplane, so feel free to take over this PR and make any changes--or I can just do it when I get back go Aarhus.

jonsterling:
Or maybe not exactly there, but somewhere.

favonia:
You are using the law of excluded nowhere...

jonsterling:
I think the problem is in `eval_nbnd`; somehow the binding is getting destroyed.

jonsterling:
Val.act is wrong!!!!!

jonsterling:
fixed now

favonia:
LOL I was so owned.

favonia:
Wait, it won't compile now.

jonsterling:
sorry, fixing!

Issue #134, opened 2018-06-30
Add natural numbers and integers


favonia:
A special case of #100.

favonia:
Natural numbers done in #153 and #155.

jonsterling:
closed by #161

PR #161, opened 2018-07-13
Adding integers, seriously.

favonia:


integer.red
```diff
@@ -0,0 +1,43 @@
...
+
+let succ-pred (n : int) : Path int (succ (pred n)) n =
+  int-rec n with
+  | pos n =>
+    nat-rec n with
+    | zero => lam _ -> pos zero
+    | suc n => lam _ -> pos (suc n)
+    end
+  | negsuc n => lam _ -> negsuc n
+  end
```
jonsterling: these examples came out nice!

equivalence.red
```diff
@@ -0,0 +1,14 @@
+import path
```
jonsterling: can you also have `invariance.red` import `equivalence.red`?

favonia: Done.

favonia:
This will close #134.

equivalence.red
```diff
@@ -0,0 +1,14 @@
...
+
+let IsContr (C : type) : type =
+  (c : _) × (c' : _) →
+    Path C c' c
```
favonia: BTW this is actually `(c : _) × ((c' : _) → Path C c' c)` which is different from other proof assistants...

jonsterling: Fwiw I don't think there's any ambiguity compared to other proof assistants, because the second part of a sigma type doesn't have a named binder. So even in a hypothetical version of Agda that has the `(x : A) * B` notation (or just Nuprl, which does have it), you would write it this way and it wouldn't be ambiguous.

favonia: Binders are blurring my point. I think `A × B → C` is usually parsed as `(A × B) → C` not `A × (B → C)`.

favonia: Well, then I should rephrase my complaint as that I think it would be more consistent to have `A → B` be syntactically identical to `(_ : A) → B` and also `A × B` to `(_ : A) × B`?

jonsterling: If we rephrase the complaint in that way, then I don't think I agree with it :stuck_out_tongue_winking_eye:

favonia: Hmm. In your opinion, should `(a : A) * (b : B) -> C -> D` be parsed as `(a : A) * ((b : B) -> (C -> D))` or ``(a : A) * ((b : B) -> C) -> D``?

jonsterling: It has to be parsed as the former, because the function arrow associates to the right.

PR #163, opened 2018-07-14
Immortal traversal

### LINKED ISSUE #120: Code duplication in Tm?
Why can't `Tm.subst` be an instance of `Tm.traverse`, for example?

jonsterling:
Resolves #120

### DIFF
--- a/src/basis/ReaderMonad.ml
+++ b/src/basis/ReaderMonad.ml
@@ -1,0 +1,28 @@
+module type S =
+sig
+  type state
+
+  include Monad.S
+  val get : state m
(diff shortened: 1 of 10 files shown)


jonsterling:
For some reason, this is slower than what we had before. But at least it is easier to understand the code, and it's all in one place.

jonsterling:
The performance degradation is actually quite bad: for univalence example, it takes 3x longer than it did before.

jonsterling:
It could be worth trying an imperative version of this, instead of the monads.

jonsterling:
Update: it was the monad that was so catastrophic. I'm switching to an imperative implementation (blech!).

Issue #166, opened 2018-07-17
Something is still backwards in extension type application


jonsterling:
Ugh!!!

jonsterling:
done.

PR #167, opened 2018-07-17
prove iso-to-equiv

jonsterling:
Much nicer than both RedPRL and yacctt proof...

~By the way, there is still something getting reversed in extension type application (see #166); you can see it if you replace the nested line types here with 2-ary extension types. So annoying... Maybe someone who has a better sense of direction than I do could have a look.~


jonsterling:
merging right away to fix embarrassing bug :stuck_out_tongue_winking_eye:

PR #173, opened 2018-07-20
start sketching out immortal typing machine

jonsterling:


jonsterling:
I've thought about it a bit and I don't think I like the idea of using this "continuation monad" to deal with re-entrant typechecking. It seems to hide too much, and I anticipate we will have state bugs that are way too hard to debug.

My thinking is that we should instead use a menhir-style incremental interface.

jonsterling:
This is a really cool paper btw: https://arxiv.org/pdf/1807.06702.pdf

jonsterling:
These ideas are also very closely related to Huet's Constructive Engine.

PR #174, opened 2018-07-22
Refine rules for the circle.

favonia:


omega1s1.red
```diff
@@ -0,0 +1,31 @@
+import path
+import integer
+import univalence
+
+let S1-univ-cover (x : S1) : type =
+  S1-rec x with
+  | base => int
+  | loop i => UA int int succ-equiv i
```
favonia: For unknown reasons it would fail if
```
UA int int succ-equiv i
```
was replaced by
```
`(V i int int succ-equiv)
```

omega1s1.red
```diff
@@ -0,0 +1,31 @@
...
+    end
+  end
+
+let winding (l : Path S1 base base) : int =
+  coe 0 1 (pos zero) in lam i -> S1-univ-cover (l i)
+
+let three : int = pos (suc (suc (suc zero)))
+
+let winding-loop-testing : Path int three (winding (loopn three)) =
+  lam _ -> three
```
favonia: The final lemma exposes bugs in `Val.ml`, maybe.

src/core/Quote.ml
```diff
@@ -199,6 +199,17 @@ struct
...
+      | S1, S1 ->
+        Tm.make Tm.S1
+
+      | Base, Base ->
+        Tm.make Tm.Base
+
+      | Loop x0, Loop x1 ->
+        (* XXX the following line is stupid *)
+        let tx = equate_dim env (`Atom x0) (`Atom x1) in
+        Tm.make @@ Tm.Loop tx
```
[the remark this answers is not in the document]
jonsterling: heheheh

omega1s1.red
jonsterling: sounds like a bug! lol.

path.red
```diff
@@ -67,9 +67,8 @@ let symm/unit
 
 let trans/filler
   (A : type)
-  (x : A)
   (p : Line A)
-  (q : Path A (p 1) x)
+  (q : [i] A with i=0 ⇒ p 1 end)
```
jonsterling: ah, interesting! I forgot that of course one can have an extension *pretype* without a valid system. nicely done.

omega1s1.red
```diff
@@ -1,11 +1,10 @@
 import path
 import integer
-import univalence
 
 let S1-univ-cover (x : S1) : type =
   S1-rec x with
   | base => int
-  | loop i => UA int int succ-equiv i
+  | loop i => `(V i int int succ-equiv)
```
favonia: I suppose a complete solution is to have the `V` tactic... not my priority now though.

jonsterling: Yes, that'll be easy to add, and I'll do it soon if nobody gets around to it.

favonia:
The merge is not done yet.

favonia:
@jonsterling There is still bug, but maybe we should merge this before it's too late?

jonsterling:
I guess I should have tested this before I merged it, it was completely broken... I'll try and fix it.

favonia:
@jonsterling Wait, what was broken? I am curious. Apparently I missed something important.

jonsterling:
@favonia First of all, the loop case was being checked in the wrong context (`cx` instead of `cxx`); additionally, I think there was some mis-use of `Cx.def`, but I could be wrong about that.

jonsterling:
(And, in the merge, the typechecking for `S1, Base, Loop` all got deleted by accident)

favonia:
Hmm well I vaguely remember the `cxx` bug was fixed but maybe there are multiple instances (!).

favonia:
Let me revive this branch to check. I think I know the actual mistake I made now.

src/core/Typing.ml
```diff
@@ -609,7 +615,7 @@ struct
       let cxx, x = Cx.ext_dim cx ~nm:nm_loop in
       let cxx_loop = Cx.def cxx ~nm ~ty:s1 ~el:(Eval.make @@ V.Loop x) in
       let mot_loop = Cx.eval cxx_loop mot in
-      let val_loopx = check_eval cx mot_loop lcase in
+      let val_loopx = check_eval cxx mot_loop lcase in
```
favonia: So theoretically it was fixed here...

favonia: Damn it the merge ate it up.

PR #176, opened 2018-07-23
Migrate to dune

### LINKED ISSUE #175: Update build to use Dune
(Dune is the new version / brand / release of jbuilder.)

Looks like a good update; eventually, dune will actually write our `.opam` files for us automatically, though this doesn't appear to be supported yet.

jonsterling:
Close #175 

Dune reports more warnings than jbuilder did (maybe that was a bug?), so now I fixed a bunch of warnings that it emitted.

I think in order to get your local environment updated, you just run `opam upgrade` and follow your nose. let me know if there are any problems.


favonia:
> Dune reports more warnings than jbuilder did (maybe that was a bug?), so now I fixed a bunch of warnings that it emitted.

It's not really any bug. Dune is the new name of jbuilder.

jonsterling:
@favonia Haha, what I meant is that maybe it was a bug that `jbuilder` was not emitting the same number of warnings as `dune` (the new version of `jbuilder`) does now.

PR #177, opened 2018-07-24
experimental fitch-style modal calculus (WIP)

jonsterling:


jonsterling:
It looks like the behavior of the `prev` operator forces me to implement the typechecker in the other direction; rather than working my way up through the spine, I must work my way *down*. This is because there are context effects in the premises of the synthesis rules.

jonsterling:
Almost done; I guess now I also have to reverse the quotation of spines to make `fix` work.

jonsterling:
Next step is to move to the version where `dfix` is a line constructor, between the fixed point and its unfolding.

src/core/Cx.ml
```diff
@@ -0,0 +1,249 @@
...
+
+let lookup i {hyps; _} =
+  let {classifier; locks; killed} = List.nth hyps i in
+  if (killed || locks > 0) && classifier != `I then
+    failwith "Hypothesis is inaccessible (modal, taste it!)"
+  else
+    classifier
+
+let lookup_constant nm tw cx =
+  if cx.all_locks > 0 then
```
jonsterling: This may not be necessary; i guess somehow every constant is a global element, and might as well have been defined inside the box modality.

jonsterling: Well, it's not quite as simple as that, but there is probably a story.

PR #179, opened 2018-07-25
Semantic domain updates

### LINKED ISSUE #178: Move definition of semantic domain into its own file
One of the things that is preventing us from fixing the gnarly modularization is that the types of the semantic domain are tied up with the implementation of the evaluator. This is a little tricky to decompose, but it seems to be a prerequisite for #162.

jonsterling:
(This is currently based on #177 because I want to avoid merge conflicts)

Something interesting I found was that the imperative update in the semantic domain for the explicit substitutions was actually not at all helping performance, and may have even slightly degraded performance.

Close #178


jonsterling:
Will just make these changes part of #177

PR #183, opened 2018-07-27
simplify domain and fix bugs

jonsterling:
it seemed like I needed this before, but now I don't.

### DIFF
--- a/src/core/Val.ml
+++ b/src/core/Val.ml
@@ -1201,6 +1201,9 @@ struct
             let phi = I.equate ri r'i in
             let phi0 = I.cmp (I.equate (`Atom x) `Dim0) phi in
             let yi, el = Abs.unleash absi in
+
+            (* If it helps, this right here is the source of the error that appears in omega1s1. - JMS *)
+            (* Format.eprintf "Here's a bad argument to rigid_vproj: %a@.@." pp_value el; *)
             Abs.bind yi @@ rigid_vproj x ~ty0:(Value.act phi0 ty0) ~ty1:(Value.act phi ty1) ~equiv:(Value.act phi0 equiv) ~el
           in
           Option.filter_map force_abs_face [face0; face1] @ List.map face sys
--- a/src/core/Val.ml
+++ b/src/core/Val.ml
@@ -1599,15 +1599,15 @@ struct
     | Tm.Var info ->
       let tty, tsys = Sig.lookup info.name info.twin in
       let rho' = Env.clear_locals rho in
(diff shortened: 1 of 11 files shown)


src/core/Val.ml
```diff
@@ -1208,6 +1201,9 @@ struct
             let phi = I.equate ri r'i in
             let phi0 = I.cmp (I.equate (`Atom x) `Dim0) phi in
             let yi, el = Abs.unleash absi in
+
+            (* If it helps, this right here is the source of the error that appears in omega1s1. - JMS *)
+            (* Format.eprintf "Here's a bad argument to rigid_vproj: %a@.@." pp_value el; *)
```
jonsterling: @favonia I don't know if this helps, but this is the call to `rigid_vproj` that causes the last example in omega1s1 to fail.

src/core/Val.ml
```diff
@@ -1199,12 +1199,8 @@ struct
           let face =
             Face.map @@ fun ri r'i absi ->
             let phi = I.equate ri r'i in
-            let phi0 = I.cmp (I.equate (`Atom x) `Dim0) phi in
             let yi, el = Abs.unleash absi in
-
-            (* If it helps, this right here is the source of the error that appears in omega1s1. - JMS *)
-            (* Format.eprintf "Here's a bad argument to rigid_vproj: %a@.@." pp_value el; *)
-            Abs.bind yi @@ rigid_vproj x ~ty0:(Value.act phi0 ty0) ~ty1:(Value.act phi ty1) ~equiv:(Value.act phi0 equiv) ~el
+            Abs.bind yi @@ vproj phi (I.act phi @@ `Atom x) ~ty0:(fun phi -> Value.act phi ty0) ~ty1:ty1 ~equiv:(fun phi -> Value.act phi equiv) ~el
```
jonsterling: @favonia This change makes the `omega1s1` example go through, but I don't know if it is correct. Can you have a look?

favonia: So it is basically the same issue. Thanks!

favonia: How do you feel about this new continuation-based API?

jonsterling: I think it's alright, but I don't yet have the right intuitions for it. I think generally, diagonals in the current implementation are very fucked up.

jonsterling:
@favonia By the way, I tried calculating that example with 5 instead of 2 and it took about a minute to compute.

jonsterling:
One of the huge bottlenecks is actually just `open_var,close_var`, that stupid locally nameless crap that I needed in the elaborator and the unifier. Just to recall, the reason that stuff even exists is I found it too hard to write the unifier with de bruijn indices.

favonia:
LOL

PR #187, opened 2018-07-27
Various Improvements

favonia:


src/core/Val.ml
```diff
@@ -21,7 +21,12 @@ type error =
   | RigidCoeUnexpectedArgument of abs
   | RigidHComUnexpectedArgument of value
   | RigidGHComUnexpectedArgument of value
+  | ApplyUnexpectedFunction of value
+  | ApplyUnexpectedCube of value
```
favonia: Not sure if `ApplyUnexpectedCube` is the best name....

jonsterling: It is not too bad...

jonsterling:
Let me merge this now to avoid conflicts.

PR #191, opened 2018-07-30
Fix inconsistency

jonsterling:
The rule for 'open' which deleted all locks was inconsistent; a conjectured consistent version is one which doesn't delete locks past a tick.

### DIFF
--- a/modal.red
+++ b/modal.red
@@ -41,11 +41,19 @@ let bool/constant (x : bool) : (b : □ bool) × Path _ x (b !) =
 
 let sequence : type = □ stream
 
+
+let Next (A : type) (x : A) ✓ : A =
+  x
+
 let sequence/cons (x : bool) (xs : sequence) : sequence =
   shut
     stream/cons
       (bool/constant x .0 !)
-      (λ _ → xs !)
+      (Next stream (xs !))
+      ; this is suspicious: we use this Next in order to essentially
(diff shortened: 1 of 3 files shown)


jonsterling:
I think the problem is that this "fix" breaks the coinduction examples...

modal.red
```diff
@@ -41,11 +41,19 @@ let bool/constant (x : bool) : (b : □ bool) × Path _ x (b !) =
...
+
 let sequence/cons (x : bool) (xs : sequence) : sequence =
   shut
     stream/cons
       (bool/constant x .0 !)
-      (λ _ → xs !)
+      (Next stream (xs !))
+      ; this is suspicious: we use this Next in order to essentially
+      ; weaken away the tick that we will not use, in order to get the right number
+      ; of locks deleted. But this is proof-theoretically very strange.
```
jonsterling: I think that this "fix" proves that the solution I proposed above, while consistent, disrupts normalizations. That is, the normal form of this proof should be the same as the normal form of the one that doesn't work, which means that we don't have the property that normalization preserves type.

jonsterling:
Even though this solution doesn't yet have good properties, I want to merge it because it is at least sound. I have written in a comment a sketch of the real solution.

jonsterling:
I'm going to merge this stupid thing and we'll decide in the next week what to do more generally (either fix or remove box modality)

Issue #24, opened 2018-04-23
[domain] Consider adding neutral hcom


jonsterling:
It is strange that in our semantic domain we have "non-neutral" values that inhabit neutral semantic types. Nothing seems broken, but it doesn't seem right either. A solution is to have `ncom` and `ncoe` (for "neutral" composition and coercion), analogous to `fcom`.

favonia:
I think it is time to consider this for `succ (winding l)`.

jonsterling:
Let's theorize it then!

favonia:
I would like to reopen this because I was also thinking about
```
  NHCom' : {dir : dir; ty : neu; cap : value; sys : comp_sys} -> neu
```

favonia:
And we probably also need a neutral coe whose type is a `neu`? I wonder how `cubicaltt` did this.

jonsterling:
Oh right, I see what you mean.

jonsterling:
Related to: #193

PR #192, opened 2018-07-31
correcting composition for positive types [WIP]

### LINKED ISSUE #157: Current type-checker of redtt is not happy with strict bools and nats
@jonsterling @cangiuli You can probably prove equality reflection for strict bools and nats through hcom, so the type checker is probably irritated when I use hcom of nat.

jonsterling:
To close #157, #24 

1. Compositions must be pushed into the arguments of constructors of positive types
2. We need neutral compositions now for neutral caps of positive type

do we need to do the same with coercion???

### DIFF
--- a/src/core/Val.ml
+++ b/src/core/Val.ml
@@ -462,6 +462,18 @@ struct
 
   and act_neu phi con =
     match con with
+    | NHCom info ->
+      let dir = Dir.act phi info.dir in
+      let ty = Value.act phi info.ty in
+      let sys = CompSys.act phi info.sys in
+      let cap =
+        match act_neu phi info.cap with
+        | Ret neu ->
+          reflect ty neu []
+        | Step cap -> cap
+      in
+      step @@ make_hcom dir ty cap sys
+
     | VProj info ->
       let mx = I.act phi @@ `Atom info.x in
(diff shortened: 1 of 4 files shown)


favonia:
I think we have to update `rigid_ghcom` too?

src/core/Val.ml
```diff
@@ -1351,12 +1351,9 @@ struct
     | Pi _ | Sg _ | Up _ ->
       make @@ GHCom {dir; ty; cap; sys}
 
-    | Bool | Nat | Int ->
-      cap
-
     (* `Ext _`: the expansion will stop after a valid
      * correction system, so it is not so bad. *)
-    | Ext _ | S1 | Univ _ | FCom _ | V _ ->
+    | Ext _ | Univ _ | FCom _ | V _ | S1 | Bool | Nat | Int ->
```
jonsterling: @favonia Do you think this would suffice?

jonsterling: @favonia Agreed. Maybe in the immortal version we will have constructors for `hcom, hcom*, com, com*, nhcom, nhcom*, ncom, ncom*` LOL

favonia:
To be honest I have not checked every rule carefully but this seems good enough.

PR #197, opened 2018-08-03
Inductive types [WIP]

jonsterling:


jonsterling:
To make the elaboration work, I really need to improve the way that the `Contextual` monad works.

jonsterling:
Ach, I might reverse my last decision, as it would complicate the account of *indexed* inductive types.

src/frontend/Elaborator.ml
```diff
@@ -149,6 +160,30 @@ struct
     | E.Quit ->
       M.ret env
 
+  and elab_datatype env edesc =
+    traverse (elab_constr env) edesc
+
+  and elab_constr env (clbl, constr) =
```
jonsterling: This is very broken, it's not treating the telescope properly.

src/core/Val.ml
```diff
@@ -2260,6 +2292,28 @@ struct
     | _ ->
       raise @@ E (RecursorUnexpectedArgument ("natural numbers", scrut))
 
+  and elim_data dlbl mot scrut clauses =
+    match unleash scrut with
+    | Intro (clbl, vs) ->
+      let _, nclo = List.find (fun (clbl', _) -> clbl = clbl') clauses in
+      (* Wonder if this is backwards ??? *)
+      inst_nclo nclo vs
```
jonsterling: This is wrong, we need to fill in the induction hyps!!

PR #210, opened 2018-08-09
miscellaneous fixes

jonsterling:


hedberg-wip.red
```diff
@@ -0,0 +1,43 @@
+import bool
+
+;; WIP
+
+data void where
```
favonia: This is so Haskell!

jonsterling:
By the way, the real challenge with improving subtyping is to carry over those improvements to the unifier-based subtyping algorithm, which is shit.

I want to have a system with just one subtyping algorithm, and it should emit constraints that are then passed to a very lightweight unifier.

favonia:
Is there an example to show off the new algorithm?

jonsterling:
@favonia Sadly any example I can think of would require unfucking the unifier-based-subtyping algorithm...

But the intention is to have better behavior wrt eta expansion, so that one doesn't need to manually eta expand elements of extension type.

PR #217, opened 2018-08-10
cleanup, fixes; close #215

### LINKED ISSUE #215: variable trouble with iterated elim
```
import path
import nat

data unit where
| triv

let Mystery (m : nat) : (n : nat) -> (Path nat m n) -> unit =
  elim m [
  | zero => lam n p -> triv
  | suc m' => lam n -> 
    elim n [
    | zero => lam p -> triv
    | suc n' => lam p -> triv
    ]
  ]
```
raises `Failed to find: m'`. The bug disappears if `Path nat m n` is replaced with something which doesn't depend either on `m` or `n`.

jonsterling:
The main change is to remove the incorrect use of unification that broke @ecavallo's code in #215; the correct code was much simpler, LOL. Thanks to @jozefg for his help in getting down with this issue.

This PR also includes random cleanup and refactoring.

Resolves #215

### DIFF
--- a/src/core/Val.ml
+++ b/src/core/Val.ml
@@ -784,54 +784,58 @@ struct
     | `Same _ ->
       args
 
+  (* Figure 8 of Part IV: https://arxiv.org/abs/1801.01568v3; specialized to non-indexed HITs. *)
+  and rigid_coe_data_intro dir abs ~dlbl ~clbl ~const_args ~rec_args ~rs =
+    let x, _ = Abs.unleash1 abs in
+    let desc = Sig.lookup_datatype dlbl in
+    let constr = Desc.lookup_constr clbl desc in
+
+    let r, r' = Dir.unleash dir in
+
+    let make_const_args dir =
+      multi_coe Env.emp dir (x, constr.const_specs) const_args
+    in
+
+    let coe_rec_arg dir _arg_spec arg =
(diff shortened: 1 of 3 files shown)


jozefg:
*taste it reaction*

src/core/Val.ml
```diff
@@ -784,54 +784,58 @@ struct
...
-    let x, tyx = Abs.unleash1 abs in
-    match unleash tyx, unleash el with
-    | Data dlbl, Intro info ->
-      (* Figure 8 of Part IV: https://arxiv.org/abs/1801.01568v3; specialized to non-indexed HITs. *)
-      let desc = Sig.lookup_datatype dlbl in
-      let constr = Desc.lookup_constr info.clbl desc in
-      let r, r' = Dir.unleash dir in
+  (* Figure 8 of Part IV: https://arxiv.org/abs/1801.01568v3; specialized to non-indexed HITs. *)
+  and rigid_coe_data_intro dir abs ~dlbl ~clbl ~const_args ~rec_args ~rs =
+    let x, _ = Abs.unleash1 abs in
```
favonia: I do not understand this line. Why not simply get a fresh variable?

jonsterling: indeed, I think I was momentarily confused! Fixing

PR #218, opened 2018-08-10
cleanup

jonsterling:


jonsterling:
For some reason, part of the travis thing isn't working? I don't understand what is going on.

jonsterling:
@favonia It's weird that it works on one travis build and not the other... I wonder if they are having some connectivity issue. annoying

jonsterling:
Travis works now, must have been temporary.

Issue #216, opened 2018-08-10
Evaluator too slow


jonsterling:
It takes almost a second to normalize the 5th winding number; we can't blame this on the Kan operations, because this is almost instantaneous in yacctt. I think it means we need to revisit some things, and maybe use a completely different style of dimension substitution.

@favonia

favonia:
Hmm, maybe we can try to provide some unsafe operations in `IAbs` (for example `unsafeBind`)? The idea is to minimize unnecessary refreshing.

ecavallo:
`five` is actually 4, btw. The real 5 takes almost 12s!

favonia:
Sorry, I can't count. Maybe I was brainwashed to believe that natural numbers start at `one`.

jonsterling:
@favonia That might help a little bit. To try it out, I tried turning off freshening completely; it breaks many examples of course, but I was able to run the winding number (five or four, lol), and it shaved off around `.25` seconds from it. So IMO it is still extremely slow (we should try to make it instantaneous), but such a performance improvement is significant.

mortberg:
Wow, 12s is surprisingly slow. And fascinating that freshening didn't affect it much either. Can you profile the code to pinpoint where it is slow?

jonsterling:
For what it's worth, this is the profile of an execution with the real `5`:

https://gist.github.com/jonsterling/2945940f555a37cc9349e79def330647

I'm not exactly sure how to interpret it; for instance, that tally of seconds in the profile doesn't reach the total running time of the program, so I found it confusing. Maybe @favonia and @ecavallo can help interpret this data...

PR #224, opened 2018-08-12
Fixes

jonsterling:
This partially fixes what turned out to be a very bad idea. It's hard to explain, but let me try. The global environment is fixed for an instance of the `Val.M` functor, but values may persist from one instance of `Val.M` to another; so what may appear to be a whnf may no longer be a whnf.

I tried to work around this by forcing `act` to re-require the global environment and patch up the value, but this leads to inconsistent & incorrect behavior which is masked by other crap. The right thing to do is maintain the invariant that an element of the semantic domain is manipulated using the version of `Val.M` which it came from. Ultimately we should probably use abstract types to enforce this, but I am worried that we would hamstring our ability to refactor if we introduced that right away. Especially considering that I'd like to blow away all that architecture.

---------

It also fixes a subtle bug in `hcom` of `fhcom`.

-----------

***This also fixes the performance problems from #216. The 5th winding number computes in `.012s`, and the 50th in less than a second.***

### DIFF
--- a/src/core/Val.ml
+++ b/src/core/Val.ml
@@ -528,22 +528,8 @@ struct
           step @@ corestriction_force v
       end
 
-    | Lvl _ ->
+    | (Lvl _ | Var _ | Meta _) ->
       ret con
-
-    | Var {name; ushift; twin} ->
-      let tty, tsys = Sig.lookup name twin in
-      let rho' = {Env.emp with global = phi} in
-      let vsys = eval_tm_sys rho' @@ Tm.map_tm_sys (Tm.shift_univ ushift) tsys in
-      let vty = eval rho' @@ Tm.shift_univ ushift tty in
-      step @@ reflect vty (Var {name; ushift; twin}) vsys
-
-    | Meta {name; ushift} ->
-      let tty, tsys = Sig.lookup name `Only in
-      let rho' = {Env.emp with global = phi} in
-      let vsys = eval_tm_sys rho' @@ Tm.map_tm_sys (Tm.shift_univ ushift) tsys in
-      let vty = eval rho' @@ Tm.shift_univ ushift tty in
-      step @@ reflect vty (Meta {name; ushift}) vsys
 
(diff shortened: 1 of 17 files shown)


src/core/Val.ml
```diff
@@ -1246,44 +1246,56 @@ struct
...
       (* This serves as `O` and the diagonal face in [F]
        * for the coherence conditions in `fhcom.sys` and `s=s'`. *)
-      let hcom_template phi y_dest ty = make_hcom
-          (Dir.make (I.act phi r) y_dest) ty
-          (Value.act phi fhcom.cap) (CompSys.act phi fhcom.sys)
+      let hcom_template phi y_dest ~ty =
+        make_hcom
+          (Dir.make (I.act phi r) y_dest)
+          ty
+          (Value.act phi cap)
```
jonsterling: Here was the extremely subtle bug: I believe that `hcom_template` should be using the outer `hcom`'s cap and system, not the inner `fcom`'s cap and system.

Sorry for hiding this change amidst formatting adjustments; I made these adjustments while I was "Getting Down" with the code and trying to understand it.

I'm still not completely confident that this matches the algorithm in @cangiuli's summary of @mortberg's `hcom-in-fcom` note (and I don't have the Immortal @favonia Note on hand). But I think this is closer.

jonsterling:
This PR now includes a fix for an extremely subtle bug (?) in `hcom` of `fhcom`: the hcoms in there were using the cap and system of the inner `fhcom`s rather than those of the outer `hcom`.

jonsterling:
The optimizations of dimension abstraction in this PR (suggested by @favonia) make calculating the (real) fifth winding number around 40% faster on my machine. It's still extremely slow.

src/core/Val.ml
favonia: Thanks!

favonia:
I cannot stop thinking that maybe we can define `value` as `con Lazy.t` or something.

PR #226, opened 2018-08-13
HITs with declared level and kind

### LINKED ISSUE #203: large inductive types
The user should be allowed to specify the universe level at which an inductive type is defined; then we should add support for the 'crude but effective stratification' hoisting to yank the whole definition up to higher universes.

jonsterling:
Resolves #203 

You can now declare a HIT with explicit level and kind, like:

```
data foo : type kan ^1 where
| con [A : type]
```

This branch also contains some renaming and general cleanup.

### DIFF
--- a/src/core/Desc.ml
+++ b/src/core/Desc.ml
@@ -1,4 +1,4 @@
-type 'a arg_ty =
+type 'a rec_spec =
   | Self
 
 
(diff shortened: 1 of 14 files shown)


favonia:
Thinking about this, I wonder if it should be `data^1`?

jonsterling:
@favonia interesting thought...

PR #227, opened 2018-08-13
Parametric HITs [WIP]

jonsterling:


jonsterling:
This is still broken, but closer. I'm utterly owned by the complex binding structure of datatype descriptions.

jonsterling:
This is broken, but maybe not as broken as it was before... I'm still not filling in the environment properly, I think.

jonsterling:
OK, now this kind of seems to work!

jonsterling:
Well, the examples work, but there are still bugs. For instance `normalize int/set` returns some de bruijn error. I expect I am not properly adding something to the environment somewhere, maybe in quotation?

src/core/Typing.ml
```diff
@@ -833,7 +833,7 @@ let check_constr_boundary_sys cx dlbl desc sys =
         | `Same | `Indet ->
           begin
             try
-              let cx', _ = Cx.restrict cx r0 r1 in
+              let cx', phi = Cx.restrict cx r0 r1 in
```
favonia: Did you use this `phi` anywhere?

jonsterling: good catch.

By the way, i had disabled the 'unused variable as build error' thing, because it made it utterly impossible to debug `redtt` (debugging is a fancy word for commenting out lots of code, possibly making variables unused, and adding `Format.eprintf`).  But this has a negative effect on code clean-ness. I wish I knew how to make it a warning but not an error.

jonsterling:
Thank you!! That's very helpful.

ecavallo:
I think I fixed that bug (I just changed something back to the way it is in master), but please double-check me!

src/core/Typing.ml
```diff
@@ -367,7 +367,7 @@ and check_constr cx dlbl param_specs constr params tms =
     | (lbl, ty) :: const_specs, rec_specs, dim_specs, _, tm :: tms ->
       let vty = V.eval tyenv ty in
       let varg = check_eval cx vty tm in
-      let cx' = Cx.def cx ~nm:(Some lbl) ~ty:vty ~el:varg in
+      let cx' = Cx.def cx' ~nm:(Some lbl) ~ty:vty ~el:varg in
```
jonsterling: Is `cx'` being used at all in this definition (aside from just getting passed around)?

jonsterling:
@ecavallo I've tried to merge `master` now into this branch and bring it up to date.

jonsterling:
@ecavallo Do you have any examples you want to try to see how broken this is?

ecavallo:
@jonsterling Propositional truncation works, but here's something that doesn't:
```
data test (A : type) where
| one [a : A]
| two [a : A] @ i [
  | i=0 ⇒ one a
  | i=1 ⇒ one a
  ]
```
raises `Failure("eval_head: de bruijn index out of range")`

jonsterling:
@ecavallo Thanks, I knew I could count on you! :laughing: 

I'll try to track it down

jonsterling:
I think the problem with this PR (and likely, the existing code) is that we are just owned by the interaction of de bruijn indices with elaboration of HIT declarations. I will try and clean it up.

src/core/Tm.ml
```diff
@@ -10,6 +10,37 @@ type bface = (tm, btm) Desc.Boundary.face
...
+  type const_spec = [`Const of tm]
+  type rec_spec = [`Self]
+  type dim_spec = [`I]
+
+  type boundary_spec = bsys
+  type param_spec = [`Param of tm]
+
+  type constr = string * (const_spec, (rec_spec, (dim_spec, boundary_spec) telescope) telescope) telescope
+  type desc = (param_spec, constr list) telescope
+end
```
jonsterling: I'd like to use this type to avoid getting confused by binding in datatype descriptions. /cc @ecavallo @favonia

src/core/Tm.ml
```diff
@@ -10,6 +10,37 @@ type bface = (tm, btm) Desc.Boundary.face
...
+
+module NewDesc =
+struct
+  type btm =
+    | Var of int
+    | Intro of
+        {clbl : string;
+         const_args : tm list;
+         rec_args : btm list;
+         rs : tm list}
```
favonia: What is rs?

src/core/Tm.ml
```diff
@@ -10,6 +10,37 @@ type bface = (tm, btm) Desc.Boundary.face
...
+
+  type bface = tm * tm * btm
```
favonia: Warning: bface will be "some flags" (like cleanness) * tm * tm * btm in the future.

src/core/dune
```diff
@@ -1,6 +1,6 @@
 (library
  (name RedTT_Core)
  (flags
-  (:standard -w -32-26-27-37))
+  (:standard -w -32-26-34-27-37))
```
favonia: ???

jonsterling: It's one of the many warnings.

src/core/Tm.ml
jonsterling: I am deciding now, I think, that it is way too much work to maintain these separate formal boundary terms (because of elaboration), and that it will be way easier to just use ordinary terms and make sure they are of the right shape.

Please stay tuned.

src/core/Tm.ml
jonsterling: Its the dimension arguments.

src/core/Tm.ml
```diff
@@ -10,6 +10,37 @@ type bface = (tm, btm) Desc.Boundary.face
...
+  type bsys = bface list
+
+  type const_spec = [`Const of tm]
+  type rec_spec = [`Self]
+  type dim_spec = [`I]
+
+  type boundary_spec = bsys
+  type param_spec = [`Param of tm]
+
+  type constr = string * (const_spec, (rec_spec, (dim_spec, boundary_spec) telescope) telescope) telescope
```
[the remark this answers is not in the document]
jonsterling: I agree, but I'm trying to write this code without needing to rewrite in a *conceptual* (i.e. not just formal) way the entire algorithms of HITs. This is to avoid getting screwed, trying to figure out which of the many changes I did at once introduced what bug.

After parametric HITs are implemented, I definitely encourage the change that you describe.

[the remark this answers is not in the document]
favonia: In my experience, the code is almost the same (and arguably more systematic).

src/core/Tm.ml
favonia: That is good because I will request wild things later, for example, the boundaries with recursors on previously defined types!

src/core/Tm.ml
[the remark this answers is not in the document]
jonsterling: @favonia I'm starting to see your point... Would you mind proposing a revised version of the `telescope` datatype with all the constructors you describe?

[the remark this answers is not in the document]
jonsterling: The thing I don't quite grasp is, how can anything be dependent on the `Rec` parts? (I guess this is a question about both my version and your version.)

src/core/Tm.ml
jonsterling: I'm glad you agree.... Otherwise we'd have to just replicate more and more of the language except in a more complicated way. what a horrible prospect!

src/core/Tm.ml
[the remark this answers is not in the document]
jonsterling: I think the main challenge is to deal with the situation where we have the intermediate state of a datatype existing (i.e. some constructors, but no eliminators). I remember that there was question in RedPRL of how to do this---what did you end up doing?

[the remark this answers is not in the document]
favonia: @jonsterling No arguments can depend on the `Rec` part, at least not right now, though boundaries can. However, the non-dependency will be enforced by type-checking automatically.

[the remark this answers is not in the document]
favonia: > I think the main challenge is to deal with the situation where we have the intermediate state of a datatype existing (i.e. some constructors, but no eliminators). I remember that there was question in RedPRL of how to do this---what did you end up doing?

It was done by two passes. The first pass grabs the applicative structure of all constructors, and then the second pass is the real elaboration. The first pass recognizes the special keyword `self`. (I am not sure if I am answering your question.)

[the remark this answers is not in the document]
jonsterling: @favonia The new more flexible schema that you propose kind of requires a global rewrite of all the HIT code, I think. I support the idea, but I think I would need some help to get it done.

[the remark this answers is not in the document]
jonsterling: I think what I will start trying to do first is to rectify some really nasty and dumb aspects of the elaborator, which will make the task a bit easier.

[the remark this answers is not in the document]
favonia: > @favonia The new more flexible schema that you propose kind of requires a global rewrite of all the HIT code, I think. I support the idea, but I think I would need some help to get it done.

I can take a look once you have done the parameters. I believe a rewrite will also close #257 automatically.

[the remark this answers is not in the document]
jonsterling: Right now, I think I am so owned by stupid de bruijn shit that it might be simpler to do the rewrite *before* adding paremeters rather than the other way around. not sure.

Issue #220, opened 2018-08-10
Vim editor mode


jonsterling:
Let's port the RedPRL vim mode to start with! I'm adding source positions to the output, so it should be possible to reuse a lot of it.

jonsterling:
@cangiuli I started hacking on this and it almost kind of works, but for some reason it nukes my main buffer whenever I call the command to run redtt. LOL. I guess I'll leave it for someone who really knows how to Get Down with vim.

cangiuli:
Is there broken code somewhere for me to look at, or is it better to just scrap it?

jonsterling:
I'll dump my broken code into a branch

cangiuli:
We should either close or rename this. @jonsterling ?

PR #229, opened 2018-08-14
Vim plugin

cangiuli:
First crack at #220. Two things are missing relative to the RedPRL mode:
* Quickfix window and "jump to error", because I haven't tried to parse the error messages yet.
* Adjusting to actual buffer width, since redtt doesn't have a flag for that.

@jonsterling Can you confirm that this works, and then either merge or tell me what to add?


cangiuli:
P.S. Somehow nobody ever complained that the RedPRL mode clobbers register `"1`, and I never noticed until just now?! (If you check the file twice and then `p`, it pastes RedPRL output instead of the last deleted text.) So this plugin actually does one thing *better* than RedPRL's!

ecavallo:
Working for me!

jonsterling:
Sorry that I went to bed before pushing my broken code! 

> P.S. Somehow nobody ever complained that the RedPRL mode clobbers register "1, and I never noticed until just now?! (If you check the file twice and then p, it pastes RedPRL output instead of the last deleted text.) So this plugin actually does one thing better than RedPRL's!

Ha, I never noticed this!

jonsterling:
I will test this when I get to the office

cangiuli:
I just create symlinks in my `~/.vim` so that my current plugin is always in sync with the repository. I guess the "install script" would just run `cp`?

jonsterling:
@cangiuli It's a nice thought, but it really don't scale if you use multiple computers, and synchronize your `.vim` folder between machines.

PR #230, opened 2018-08-14
Rectification Campaign II: Restrictions, etc.

### LINKED ISSUE #212: Add neutral coe
The tricky part is to pass some `Up` or `neu` *under a binder*. The following seems to be the only solution if we want to reuse `IAbs`:
```
  | NCoeAtType of {dir : dir; univ : value; abs : neu_abs; el : value}

...

and neu_abs = neutral IAbs.abs
and neutral = NeuNode of {neu : neu; action : I.action}
```
However, this will lead to very ugly and clumsy code. What should we do?

### LINKED ISSUE #222: composition in path types
```
import path

let experiment (A : type) (u : A) : (p : Path A u u) -> Path (Path A u u) p p =
  lam p i -> 
    comp 0 1 p [
    | i=0 => lam j -> p
    | i=1 => lam j -> p
    ] 
```
raises `Failure("Extension subtype: nope")`. If you replace the `p`'s in the `comp` with holes, you can see something weird is happening to the extension types.

### LINKED ISSUE #228: bug in path type coercion?
```
import path

let thing (A : type) (a,b : A) (p : Path A a b) : Path (Path A a b) p p =
  λ i j →
  let ty : dim → type = λ k →
    let h : dim → A = λ l →
      comp 0 l a [
      | k=0 ⇒ λ z → p z
      | k=1 ⇒ λ z → p z
      ]
    in
    Path (Path A (h 0) (h 1)) (λ l → h l) (λ l → h l)
  in
  (coe 0 1 (λ _ z → p z) in ty) i j
```
raises
```
rigid-rigid mismatch: (hcom 0 j
                       A
                       a
                       [%6539=0 <x0> (@ p x0)] [%6539=1 <x0> (@ p x0)]) ≐
                       (@ p j) : A
Encountered error:
    Failure("rigid-rigid mismatch")
```
No clue what's going wrong, but I _think_ this should typecheck...

### LINKED ISSUE #234: variables getting reversed in HITs
```
data O where
| obase 
| oloop (y : O) @ i
  [ i=0 ⇒ y
  | i=1 ⇒ y
  ]
```
raises `Failure("check_dim_cmd: expected dimension")`, but
```
data O where
| obase 
| oloop (y : O) @ i
  [ y=0 ⇒ i
  | y=1 ⇒ i
  ]
```
typechecks. I've failed to figure out what needs to be switched around to fix it :(

### LINKED ISSUE #237: coercing in Code-like type...?
I don't know what this one is about.
```
import path

data unit where
| triv

data twonit where
| wrap [x : unit]

let Code (n : twonit) : type =
  elim n [ wrap n' => dim -> unit ]

let test (p : Path twonit (wrap triv) (wrap triv)) : dim -> unit =
  coe 0 1 (lam _ -> triv) in (lam i -> Code (p i))
```
raises
```
Encountered error:
    Tried to unleash ((up
                       (U kan 0)
                       (twonit.elim
                        <clo (U kan 0) & (up
                                          (twonit)<[i/i]>
                                          (@ p i)
                                          [[?i=0 ((wrap (triv   )  ))<[0/i] ∘ [i/i]>]
                                          [?i=1 ((wrap (triv   )  ))<[1/i] ∘ [i/i]>]])>
                        (@ p i)
                        [wrap
                         <clo (# <_> unit) & (up
                                              (twonit)<[i/i]>
                                              (@ p i)
                                              [[?i=0 ((wrap (triv   )  ))<[0/i] ∘ [i/i]>]
                                              [?i=1 ((wrap (triv   )  ))<[1/i] ∘ [i/i]>]])>])
                       [[?i=0 (# <_> unit )]
                       [?i=1 (# <_> unit )]]))<[%4732/i]> as extension type.
```
Defining `Code` as just `dim -> unit` without the `elim` undoes the bug, as does replacing `dim -> unit` with just `unit`.

### LINKED ISSUE #238: hcom in strict type finding non-intro
```
import path
import J

data vacu where
| more (x : vacu)

let Code (m : vacu) (n : vacu) : type =
  elim n [ more n' => Path vacu m n' ]

let test (m : vacu) (p : Path vacu (more m) (more m)) : Path vacu m m =
  J vacu (more m) (lam n _ -> Code m n) (lam _ -> m) (more m) p
```
raises `Failure("rigid_hcom_strict_data: peel_arg")`. The bug disappears if I make `vacu` nonrecursive.

### LINKED ISSUE #240: "Tried to evaluate non-dimension term ... as dimension"
I am getting this error when I try to eliminate out of the torus into a certain extension type. Sounds like a bug in the eliminators...

jonsterling:
+ Removing box modality, because the cure was worse than the disease. We should wait until we can implement *time warps* :wink: 

+ Preliminary work toward #231 

+ resolves #228 
+ resolves #222 
+ resolves #234  (?)
+ resolves #212 
+ resolves #237 
+ resolves #240 
+ resolves #238 
+ very basic `auto` tactic (will become better as I refactor the refiner)
+ disabling strict composition for now, because what we had was wrong, and I am not yet decided on how to make it right

Huge thanks to @ecavallo for exercising all the bugs fixed in this PR.


jonsterling:
Roadmap:

- [ ] change the elaborator to push around restrictions in the same way that the typechecker does
- [ ] hopefully, change the restriction pretypes to have an explicit intro and elim form

src/core/Val.ml
```diff
@@ -1451,7 +1451,9 @@ struct
     let rho' =
       Env.push_many
         begin
-          (* This is not backwards, FYI. *)
+          List.rev @@
+          (* ~~This is not backwards, FYI.~~ *)
+          (* NARRATOR VOICE: it was backwards. *)
```
jonsterling: @ecavallo :laughing: :sunglasses:

jonsterling: I should be forbidden to use `List.rev`... I get so owned by it.

jonsterling:
@ecavallo The next thing I need to do to make your example work is rewrite `tac_elim`. The problem is that it tries to refine the match clauses one by one, but this doesn't yield enough information to construct the boundary of each subgoal, since generally this requires one to know *all* the clauses of the elimination advance.

This is really hard, and I'm not yet sure how to completely do it; the issue is that I need to somehow unleash subgoals for all the clauses first, and then fill them in clause-by-clause. This is slightly better than where I was, but the problem is that I don't even know what type to give those subgoals.

jonsterling:
@ecavallo Can you try this out now? I think I fixed the issues you described, but given how complex the code is, I have no doubt that there's more problems.

ecavallo:
Nice! Here's something that still doesn't work:
```
let compose (o1,o2 : O) : O =
  elim o1 [
  | obase ⇒ o2
  | oloop (o1' ⇒ compose/o1'/o2) i ⇒ oloop compose/o1'/o2 i
  ]

let compose/obase/r (o : O) : Path O (compose o obase) o =
  elim o [
  | obase ⇒ lam _ -> obase
  | oloop (o' ⇒ compose/obase/r/o') i ⇒ lam j -> oloop (compose/obase/r/o' j) i
  ]

```
The second definition raises `Did not find what was expected in the environment: (up O #2 [])`. I guess it's a problem with dependent elim specifically?

jonsterling:
I see the bug, I think.

jonsterling:
This partly addresses #231 (what remains is to add explicit intro and elim forms), and really addresses a lot of general broken-ness.

The main idea of this pull request is use the judgmental structure to deal with restrictions, and push them down into subgoals asynchronously as much as possible, rather than checking them synchronously.

An effect of this is that if you have a restricted goal of sigma, pi or extension type, you can apply the introduction form, and the appropriate restrictions will appear in the subgoals. For positive types, you must satisfy the restriction immediately, so to enable a partial construction that doesn't yet satisfy this restriction, you should wrap that code in `?{ ... }`. This has the effect of unleashing an auxiliary hole to stand for what you're trying to build, and then clears out the restriction for the tactics that are inside the braces. So long as you don't satisfy the restriction, what you have constructed will not be used to fill in the hole, but you can see your progress.

jonsterling:
These changes bring us closer to being able to implement the immortal typechecking algorithm, in which you don't have to manually eta expand things of extension type in any cases.

jonsterling:
Oh, and this PR fixes the completely wrong calculation of boundaries for eliminator clauses, LOL

jonsterling:
This PR now includes a change which improves the subtyping theory for extension types; the idea is to use my immortal idea of continuing *checking* neutrals at negative type, delaying the mode switch until we hit a positive type.

This is not the complete solution yet, because we later need to actually *define* subtyping for negative types in terms of this.

favonia:
This should work:
```
let ncoe/int (n : int) : Path int (coe 0 1 n in λ _ → int) n = lam _ -> n
```

jonsterling:
I think I will actually merge this now that it is cleaned up; the PR is getting too big.

Issue #154, opened 2018-07-11
Quote.generic should return the extended env too.


favonia:
Right now, extending environments and getting new variables are two different operators, which opens the door to various bugs. There should be a single operator to do both.

favonia:
Will be closed by #270.

PR #242, opened 2018-08-17
Fixing general backwardness

### LINKED ISSUE #236: some kind of binding trouble
Just by adding another dimension variable (`w`) to a connection, I get `Failure("rigid-rigid mismatch")`:
```
let connection/and/more
 (A : type)
 (p : dim → A)
 : [i j w] A [
   | j=0 ⇒ p 0
   | i=0 ⇒ p 0
   | j=1 ⇒ p i
   | i=1 ⇒ p j
   ]
 =
 λ i j w →
   let face : dim → dim → A =
     λ l k →
       comp 1 l (p k) [
       | k=0 ⇒ λ _ → p 0
       | k=1 ⇒ p
       ]
   in
   comp 0 1 (p 0) [
   | i=0 ⇒ face 0
   | i=1 ⇒ face j
   | j=0 ⇒ face 0
   | j=1 ⇒ face i
   | i=j ⇒ face i
   ]
```
(You can delete all but one of the faces in the type and definition and get the same result.)

jonsterling:
The point of this PR is to fix a ton of things that were mutually backwards, reversed, in a horrible nest of confusion.

Resolves #236 

The method by which I am doing this is to unify the basic data structures under the metaphor of the context and environment extending toward the right, and being read from right to left (like on paper!). Most of the bugs had to do with how to extend a context or environment by multiple items; if you are using lists for the context, then you must *reverse* the things before you put them in. But if you are using `bwd` for the context and `list` for the things you are adding (the Conor McBride yoga), then, you don't consciously need to do any reversing. We have the "fish combinator" `<>< : bwd -> list -> bwd`, which is meant to handle just this kind of case.

More than a decade ago, Conor realized I guess that using forward lists for everything just completely owns you when trying to write a proof assistant. We should really just ban `List.rev`.


jonsterling:
Enough is fixed that now @ecavallo's broken example works:

```
let test (A : type) (a : A) : [i j] A [i=0 ⇒ a | i=1 ⇒ a] =
 λ i j →
 comp 0 1 a [
 | i=0 ⇒ λ _ → a
 | i=1 ⇒ λ _ → a
 ]
```

favonia:
Perhaps we are in a better shape to address #154 now.

PR #245, opened 2018-08-17
Hedberg

### LINKED ISSUE #207: Z is a set
When trying out my problem we will need a proof that the integers form a set.  We can either prove this directly using J or using Hedberg. Having both proofs could be interesting just to compare for efficiency.

For proofs in cubicaltt/hcomptrans see setZ' (using J) and ZSet (using Hedberg) in: https://github.com/mortberg/cubicaltt/blob/pi4s3/examples/brunerie2.ctt

### LINKED ISSUE #239: Constant square in redtt
Here's a small puzzle that we will need to solve by next week: given `p : Path A a a` construct a square with `p` everywhere on the boundary.

In cubicaltt this uses lots of connections/reversals:

https://github.com/mortberg/cubicaltt/blob/master/examples/prelude.ctt#L165

We would definitely want to avoid just copying this as it will be pretty huge without native connections/reversals. Can someone come up with a direct definition like the one used for connections?

I don't think it's important what the diagonal is, but if you can get it to be `p` as well then you get a bonus point!

ecavallo:
* Prove that `nat` and `int` are sets using Hedberg's theorem (resolves #207)
* Improve connection efficiency and define "constant square" (resolves #239)
* Make a folder for examples.

The proof of Hedberg is different from the one in cubicaltt, though the idea is the same. I did use some connections, so there may be a more efficient way.


example/hedberg.red
```diff
@@ -0,0 +1,63 @@
...
+    | i=0 ⇒ λ k →
+      st (p j) (p j)
+         (neg/is-prop-over (λ j → neg (Path A (p j) (p j)))
+           (λ c → c (square 0))
+           (λ c → c (square 1))
+           j)
+         k
+    | i=1 ⇒ λ _ → q j
+    | j=0 ⇒ λ k → connection/or A (cap 0) i k
+    | j=1 ⇒ λ k → connection/or A (cap 1) i k
```
mortberg: One thing to consider if you take partial elements seriously would be to allow sharing of branches in systems. That way the two last cases could be written:
```
    | j=0 \/ j=1 ⇒ λ k → connection/or A (cap j) i k
```
This happens quite often and it might help with memory consumption.

jonsterling: That'd be so cool! Let's talk about it at dagstuhl.

mortberg:
Looks good to me! Two minor results that would be easy to add is that `void` and `unit` are h-propositions.

example/hedberg.red
jonsterling: I think i see how to make that work...

PR #248, opened 2018-08-18
direct proof that IsEquiv f is a prop

### LINKED ISSUE #246: Implement better PropIsEquiv
One problem I had in cubicaltt was that the normal form of the proof of PropIsEquiv: 

https://github.com/RedPRL/redtt/blob/master/example/univalence.red#L113

was huge and hence had to be made opaque in order to not blow up the typechecker. It seems like the same is true for redtt:

https://github.com/RedPRL/redtt/commit/b99bac4eb7a25956403afb6890bac69720e65b01#commitcomment-30139222

I ported a much more direct proof to yacctt from cubicaltt:

https://github.com/mortberg/yacctt/commit/dd238a28f4710485cfc3a399f7ed3fea48239ef1#diff-4597bedd84ee09b2a30c44d58082463bR279

Note that I'm doing an hcom in Fiber which probably can be "simplified" even further:

https://github.com/mortberg/yacctt/commit/dd238a28f4710485cfc3a399f7ed3fea48239ef1#diff-4597bedd84ee09b2a30c44d58082463bL293

The normal form of these proofs are *much* smaller and they compute a lot better and there is probably no need to make them opaque during TC. However I conjecture that something like Evan's recent connection trick can be used to find an even simpler proof in redtt.

ecavallo:
To close #246.

I didn't see any obvious opportunities to improve on the yacctt algorithm. The normal form of PropIsEquivDirect is a few hundred lines shorter than that of PropIsEquiv (~750 vs ~1150).

### DIFF
--- a/example/univalence.red
+++ b/example/univalence.red
@@ -113,6 +113,33 @@ opaque
 let PropIsEquiv (A,B : type) (f : A → B) : IsProp (IsEquiv A B f) =
   λ e0 e1 i b → PropIsContr (Fiber A B f b) (e0 b) (e1 b) i
 
+; A direct proof that IsEquiv f is a prop, ported from cubicaltt to yacctt to redtt
+let PropIsEquivDirect (A,B : type) (f : A → B) : IsProp (IsEquiv A B f) =
+  λ ise ise' i y →
+    let a : A = (ise y).0.0 in
+    let p : Path B (f a) y = (ise y).0.1 in
+    let c : (w : Fiber A B f y) → Path (Fiber A B f y) w <a,p> =
+      (ise y).1
+    in
+    let a' : A = (ise' y).0.0 in
+    let p' : Path B (f a') y = (ise' y).0.1 in
+    let c' : (w : Fiber A B f y) → Path (Fiber A B f y) w <a',p'> =
+      (ise' y).1
+    in
+    < c' <a , p> i
+    , λ w →
+        let sq : PathD (λ j → Path (Fiber A B f y) w (c' <a,p> j)) (c w) (c' w) =
+          λ i j →
+            comp 0 1 (connection/and (Fiber A B f y) (c' w) i j) [
+            | i=0 ⇒ λ k → connection/and (Fiber A B f y) (c w) k j
+            | i=1 ⇒ λ _ → c' w j
+            | j=0 ⇒ λ _ → w
+            | j=1 ⇒ λ k → c' (c w k) i
+            ]
+        in
+        sq i
+    >
+
 opaque
 let EquivLemma
   (A,B : type) (E0, E1 : Equiv A B)


mortberg:
Which definition of connection/And is this using? The new one @ecavallo found earlier today? 

Did this help with typechecking the univalence proof without opaque?

I think the hcom in Fiber might unfold to something suboptimal unless we define com in Path directly (i.e. not using an hcom of coe's). Did you implement this optimization @jonsterling ? This might be a good place to test it to see how much shorter the normal form gets with or without it.

Finally; maybe we should add an option to print the number of hcom and coe in the normal form of a term as this is a good estimate for how computationally complex they are (as opposed to just staring at a wall of text or comparing number of lines/characters)?

jonsterling:
@mortberg I did try defining `com` in several connectives directly and it actually had no effect on performance, might have made it a little worse, surprisingly. But I did not check the size of the normal forms, so this is worth doing: and if I turn off `opaque`, this change might indicate a big difference.

So, let me try that optimization again :smile:

example/univalence.red
```diff
@@ -113,6 +113,33 @@ opaque
...
+        let sq : PathD (λ j → Path (Fiber A B f y) w (c' <a,p> j)) (c w) (c' w) =
+          λ i j →
+            comp 0 1 (connection/and (Fiber A B f y) (c' w) i j) [
+            | i=0 ⇒ λ k → connection/and (Fiber A B f y) (c w) k j
+            | i=1 ⇒ λ _ → c' w j
+            | j=0 ⇒ λ _ → w
+            | j=1 ⇒ λ k → c' (c w k) i
+            ]
+        in
+        sq i
```
jonsterling: Is this cut included here for clarity? Otherwise, I think it would be possible to just do the `lam j -> comp...` directly. But if you feel it's more clear this way, I'm fine with it.

ecavallo: Just clarity.

jonsterling: cool!

jonsterling:
BTW, this pull request exposes a bug in `redtt`: if I try to use `PropIsEquivDirect` in place of `PropIsEquiv`, I get the following error:

```
Encountered error:
    (car (car ((cdr E0) (up (up (U kan 0) B []) #1 []))))
    ≠
    (ncoe 0 %18768
     <%18773> ((up (U kan 0) A []))<%18772 <-> %18773 ∘ %18766 <-> %18772 ∘ [0/i] ∘ [0/i] ∘ i <-> i ∘ [0/i] ∘ %18765 <-> %18768 ∘ [0/i] ∘ %18765 <-> %18766 ∘ %18763 <-> %18765 ∘ %18761 <-> %18763 ∘ %18738 <-> %18761 ∘ [0/i] ∘ [i/i] ∘ [i/i] ∘ [%18738/j]>
     ((up
       (up (U kan 0) A [])
       (car (car ((cdr E0) (up (up (U kan 0) B []) #1 []))))
       []))<[0/i] ∘ [0/i] ∘ i <-> i ∘ [0/i] ∘ %18765 <-> %18768 ∘ [0/i]>)
```

If I make `PropIsEquivDirect` opaque, then it of course works; this suggests to me that there is a bug either in evaluation or quotation. (It shouldn't block this PR.)

PR #250, opened 2018-08-18
Coersion in strict data types and Omega1(S1).

### LINKED ISSUE #241: coe in discrete types on neutral elements
This should work:
```
let ncoe/int (n : int) : Path int (coe 0 1 n in λ _ → int) n = lam _ → n
```

favonia:
Close #241.

### DIFF
--- a/example/omega1s1-wip.red
+++ b/example/omega1s1-wip.red
@@ -42,3 +42,28 @@ let winding-loopn (n : int) : Path int (winding (loopn n)) n =
     | suc (n ⇒ loopn) ⇒ λ i → pred (loopn i)
     ]
   ]
+
+
+; speed test
+
+let two : int = pos (suc (suc zero))
+
+let winding-loop-testing0 : Path int two (winding (loopn two)) =
+  auto
+
+let nat/five : nat =
+  suc (suc (suc (suc (suc zero))))
+
+let nat/25 : nat =
+  plus nat/five (plus nat/five (plus nat/five (plus nat/five nat/five)))
+
+let int/50 : int =
+  pos (plus nat/25 nat/25)
+
+let int/5 : int = pos nat/five
+
+let winding-test/5 : Path int int/5 (winding (loopn int/5)) =
+  auto
+
+let winding-test/50 : Path int int/50 (winding (loopn int/50)) =
+  auto
--- a/example/omega1s1-wip.red
+++ b/example/omega1s1-wip.red
@@ -17,37 +17,28 @@ let loopn (n : int) : Path s1 base base =
   | pos n ⇒
     elim n [
     | zero ⇒ auto
-    | suc (n ⇒ loopn) ⇒ trans s1 (λ i → loop i) loopn
+    | suc (n ⇒ loopn) ⇒ trans s1 loopn (λ i → loop i)
     ]
   | negsuc n ⇒
     elim n [
     | zero ⇒ symm s1 (λ i → loop i)
-    | suc (n ⇒ loopn) ⇒ trans s1 (symm s1 (λ i → loop i)) loopn
+    | suc (n ⇒ loopn) ⇒ trans s1 loopn (symm s1 (λ i → loop i))
     ]
   ]
 
 let winding (l : Path s1 base base) : int =
   coe 0 1 (pos zero) in λ i → s1-univ-cover (l i)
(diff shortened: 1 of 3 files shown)


src/core/Val.ml
```diff
@@ -1358,15 +1360,14 @@ struct
...
+    | Pi _ | Sg _ ->
       make @@ GHCom {dir; ty; cap; sys}
 
     (* `Ext _`: the expansion will stop after a valid
-     * correction system, so it is not so bad. *)
-    | Ext _ | Univ _ | FHCom _ | V _ | Data _ ->
+     * correction system, so it is not so bad.
+     *
+     * `Up _`: perhaps we can have nghcom? *)
+    | Ext _ | Univ _ | FHCom _ | V _ | Data _ | Up _ ->
```
favonia: We need a neutral hcom or a neutral ghcom so that a potential eliminator knows how to deal with it.

example/omega1s1-wip.red
```diff
@@ -17,37 +17,28 @@ let loopn (n : int) : Path s1 base base =
...
-let int/50 : int =
-  pos (plus nat/25 nat/25)
-
-let int/5 : int = pos nat/five
-
-let winding-test/5 : Path int int/5 (winding (loopn int/5)) =
-  auto
-
-let winding-test/50 : Path int int/50 (winding (loopn int/50)) =
-  auto
```
favonia: Does anyone miss this? I can put it back...

jonsterling: Let's keep it, just to make sure we do my introduce some really bad performance degradation. We could even move to different file if you felt it was clutter

favonia: Done.

jonsterling:
@favonia when you run the 50th winding number on your machine, what happens? Travis seems to time out after 10 minute :frowning:

favonia:
Glad that we kept the speed test. I apparently screwed up something.

jonsterling:
@favonia Laziness is notoriously difficult to predict :laughing: I wonder if some change caused a bad thunk to get forced...

favonia:
@jonsterling I am afraid this PR exposes some ugly existing performance bugs. If you tried the original example file, it is totally fine. The fact that I proved something before the ultimate speed testing halts everything.

favonia:
Sorry. Please forget about the above comment (if it still showing in your browser). I probably found the actual cause.

example/omega1s1-wip.red
```diff
@@ -17,18 +17,35 @@ let loopn (n : int) : Path s1 base base =
   | pos n ⇒
     elim n [
     | zero ⇒ auto
-    | suc (n ⇒ loopn) ⇒ trans s1 (λ i → loop i) loopn
+    | suc (n ⇒ loopn) ⇒ trans s1 loopn (λ i → loop i)
```
favonia: This line and the similar line below seem to be causing the trouble.

jonsterling: Heh.. So the issue is that when it associates in one way it is fast, and in the other way infeasible? That's very interesting.

favonia: It somehow makes sense. I changed the order to make the lemma easier to prove. The new one produces `isuc` from the outside and is perfect for induction.

jonsterling: Shall we try a harder version of the lemma? I guess analogous to tail recursive function

ecavallo: This performance asymmetry also exists in cubicaltt, btw.

jonsterling: I guess to prove the "fast" version, it will take some gnarly coercion. Sadly, the goals are tens of thousands of lines long, so it's a little hard to tell what needs to be done.

ecavallo: @favonia Can you elaborate on why it makes sense? I guess the performance of coe in fcom depends more on the complexity of the tubes than that of the cap?

favonia: @ecavallo I am not sure. I have to think harder.

favonia:
@jonsterling BTW, since we have parameters now, are you treating something like `list` as strict? I might have treated `list s1` wrongly as a strict type.

favonia:
We decided to give up the speed test now.

PR #252, opened 2018-08-18
prove UA (IdEquiv A) is constant A

### LINKED ISSUE #249: V-types puzzle
A while ago I asked whether we can prove the following fact about V-types:
```
goal (A : U) : Path (Path U A A) (<i> V i A A (idEquiv A)) (<_> A) =
```

A solution was proposed by Jon using a connection: https://github.com/RedPRL/redtt/blob/master/example/univalence.red#L246

This is essentially `<j i> V (i /\ j) A A (idEquiv A)` which gives us a path with the right endpoints (except it's reversed), but @favonia pointed out to me just now that this is not a path over refl as desired in the original formulation of the puzzle. So the puzzle is still open. 

This puzzle is interesting as it is part of the definition of the Hopf fibration for S^2 defined as a 2-HIT. This is the part of the computation of Brunerie's number where I'm stuck so we will need to figure out how to solve the puzzle (or circumvent it) at some point if we want to try my problem in redtt.

For reference, here's the proof in cubicaltt: https://github.com/mortberg/cubicaltt/blob/pi4s3/examples/problem.ctt#L288

Note that it uses a 2-d Glue type... This is something that I cannot see how to do directly using V-types.

ecavallo:
To resolve #249. I don't know what efficiency improvements are possible...

### DIFF
--- a/example/retract.red
+++ b/example/retract.red
@@ -5,6 +5,7 @@ let Retract (A,B : type) (f : A → B) (g : B → A) : type =
   (a : A) →
     Path A (g (f a)) a
 
+; Adapted from https://github.com/HoTT/book/issues/718
 let path-retract/preserves/refl (A : type) (R : A → A → type)
   (s : (x,y : A) → (R x y) → Path A x y)
   (r : (x,y : A) → (Path A x y) → R x y)
--- a/example/retract.red
+++ b/example/retract.red
@@ -1,0 +1,58 @@
+import path
+import connection
+import equivalence
+
+let Retract (A,B : type) (f : A → B) (g : B → A) : type =
+  (a : A) →
+    Path A (g (f a)) a
+
+let path-retract/preserves/refl (A : type) (R : A → A → type)
+  (s : (x,y : A) → (R x y) → Path A x y)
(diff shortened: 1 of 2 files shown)


jonsterling:
pretty cool!!

example/retract.red
```diff
@@ -0,0 +1,57 @@
...
+     | k=1 ⇒ λ _ → x
+     ]
+  in
+  λ i j →
+    comp 0 1 (cap2 i j) [
+    | j=0 ⇒ λ _ → x
+    | j=1 ⇒ face i
+    | i=0 ⇒ λ _ → q j
+    | i=1 ⇒ face j
+    ]
```
mortberg: Wow, this was quite complicated..

ecavallo:
Unfortunately it seems that trying to normalize it triggers some bug in redtt...

ecavallo:
The retract proof is from https://github.com/HoTT/book/issues/718, by the way (added comment). I restricted it to just proving that the roundtrip preserves refl, as I think the general case would need more compositions.

mortberg:
It would be very interesting to see how big the normal form is. My experience with normal forms of univalence is not very pleasant...

PR #256, opened 2018-08-21
Pointed types

cangiuli:


jonsterling:
@ecavallo that sounds like the kind of thing that might prevent self-own

favonia:
I think the general rule is that, do not start lambdas for extension types unless you have done all the case splitting.

PR #262, opened 2018-08-23
Debugging, cleanup

jonsterling:


src/core/Restriction.ml
```diff
@@ -5,12 +5,18 @@ type dim = I.t
 
 type eqn = dim * dim
 
-module UF = DisjointSet.Make (PersistentTable.M)
```
jonsterling: @favonia I don't know if the changes in this file are useful at all. should we revert or keep them?

jonsterling:
btw don't worry about this PR, it's most garbage experimentation, but I will extract some things from it ultimately.

example/torus.red
```diff
@@ -21,7 +21,7 @@ let t2c (t : torus) : s1 × s1 =
   | pt ⇒ <base, base>
   | p/one i ⇒ <loop i, base>
   | p/two i ⇒ <base, loop i>
-  | square i j ⇒ refl
```
cangiuli: :laughing:

jonsterling: taste it

jonsterling: It's a good test, but it is a little confusing in context.

Issue #231, opened 2018-08-14
Find a way to limit the pain of restriction pretypes (or remove them)


jonsterling:
The restriction pretypes are brutal, and a lot of things are broken because of them. I would like to find a way to remove them, or at least limit their influence. Let me start by recalling why they exist in the first place: we need to be able to have a goal with a restriction, and then skolemize that goal as some kind of type (or other structure) that can be made into a hole in the global context.

Unfortunately they make the implementation very complicated, and in practice really unreliable and confusing. In most places, we are either too clever wrt restrictions or not clever enough.

I would like to remove them and instead make restrictions part of the *judgments*. Stratifying things in this way would clarify a LOT. I don't yet see how to remove the restriction pretypes entirely without making holes impossible (there is probably a way), but to start, I think we could try and implement this judgmental structure.

/cc @ecavallo

jonsterling:
Another thing is that if we use judgmental structure for these in the main (and only internalize them as pretypes when actually necessary, such as skolemizing a goal) it would be more clear how to give an explicit introduction and elimination form to the restriction type, and avoid the question of subtyping between a restriction type and a non-restriction type entirely.

favonia:
Yes, please.

favonia:
Yes! It is weird that restriction types are not internalizing something judgmental!

jonsterling:
@favonia 
> Yes! It is weird that restriction types are not internalizing something judgmental!

Indeed! I think that general principle, that types should internalize judgments, is yet another Law of Eliminating Mortality...

PR #269, opened 2018-08-29
use closures for extension type lambdas; replace Tm.Rst with Tm.Ext; renaming

### LINKED ISSUE #244: Rename everything that has to do with "restriction"
This name is so confusing!! Let me count the ways:

1. We have something currently called "co-restriction", which is basically `(r = r') -> A`; this should actually be called "restriction".

2. What is currently called restriction is, conceptually, the pretype of extensions of a partial element to a total element. But we can't use "extension" for this, that's already taken.

I have proposed calling it "boundary refinement" and "boundary refinement types". Other ideas are welcome.

/cc @ecavallo @mortberg @cangiuli

jonsterling:
@mortberg @cangiuli 

Also resolves #244 and #231


jonsterling:
@mortberg It's slightly faster with the new code, but not by a huge amount.

jonsterling:
This should be ready shortly, I just have to check that I haven't disrupted the interactive side of things.

PR #265, opened 2018-08-28
Dimension equation theory oracle

favonia:
I need to commit this before switching back to the master branch.

PR #270, opened 2018-08-30
New redtt core

jonsterling:
I'm sketching some ideas for how to formulate the new evaluator in a cleaner and more systematic way, which should also enable some optimizations to avoid unnecessary swapping, etc.


jonsterling:
Ahh, I think I realized that the idea of naive / syntactic substitution is just wrong in this setting, even if followed by 're-evaluation'.

Consider `x = y !!- loop[x] <0/y>`; with the naive version, this remains `loop[x]`, and does not become `base`.

I guess the meaning of substitution really has to be, add the equation to the theory and re-evaluate or something.

@favonia Do you have any clarifying Thoughts here?

favonia:
The substitution is done to both the term and the oracle. That's why we have the two-layer oracle to deal with hiding. The semantics in terms of stratified judgments is
```
dimension variables; equation theory; expression variables |- term : type
```
so any dimension substitution applies to the rest of the judgment.

In this case. You should have `x=0 !!- loop[x]` after substitution.

jonsterling:
@favonia I see, now I think I remember.

jonsterling:
@favonia Does the following rule make sense? In particular the third premise to calculate the substitution. (I'm checking whether I understood your comment.) 

screen shot 2018-08-30 at 11 56 53 am

favonia:
Oh, I have not updated the notes. I guess `alpha` should just join the oracle.

favonia:
Oops. I forgot `alpha` is actually a list. This notation makes some sense then.

favonia:
I now implemented the subst and swap operators directly in #265 to save your brain power.

favonia:
> @favonia Does the following rule make sense? In particular the third premise to calculate the substitution. (I'm checking whether I understood your comment.)

I fixed the rule in the LaTeX note.

favonia:
> Oops I guess I also need to implement apartness. (It can be simulated by try union r r' h; false with _ -> true but this is so ugly.)

Done.

> Oops I guess I also need to implement apartness. (It can be simulated by try union r r' h; false with _ -> true but this is so ugly.)

Done in #265.

favonia:
@jonsterling I think `equate` is a good name. How about me doing commit something to #265 so that you can cherry-pick? Also, do you want a pretty printer? If so, is the following format readable?
```
x~0,y~0,z~#7,w~#6,s~#7,t~#7
```
The above line means `x=y=0` and `z=s=t`. `#6` and `#7` have no meaning outside the oracle.

jonsterling:
@favonia I already brought in the files, so technically there is no need to cherry-pick --- except the git history is not correct, since it does not indicate that you wrote those files. Please feel to make changes directly to this branch, so that we avoid needless merge conflicts.

Also feel free to make any renamings you like :)

jonsterling:
And the pretty printing format you described sounds good!

favonia:
> And the pretty printing format you described sounds good!

Done (modulo bugs).

favonia:
> I already brought in the files, so technically there is no need to cherry-pick

I think you can cherry-pick a commit out of the history, but oh well it is done now.

src/core/NewRestriction.ml
```diff
@@ -0,0 +1,170 @@
...
+module T = PersistentTable.M
+
+type t =
+  {index : (atom, int) T.t;
+   next : int;
+   rank : (int, int) T.t;
+   mutable parent : (int, cls) T.t}
+
+let emp () =
+  let size = 100 in
```
favonia: I might have picked the wrong initial size...

src/core/NewDomain.ml
```diff
@@ -482,7 +483,7 @@ struct
 
     | Neu info ->
       let neu = {head = NCoe {r; r'; ty = Abs (x, info.neu); cap}; frames = Emp} in
-      let ty = Val.hsubst r' x rel tyx in
+      let ty = val_hsubst r' x (Rel.hide' x rel) tyx in
```
favonia: @jonsterling I believe there was a bug when `Rel.hide` was missing. (Unless you claimed `x` is always fresh.)

jonsterling: SHould that just be built into Val.hsubst?

favonia: No, because in general `subst r x (hide x rel)` is different from `subst r x rel`.
```
subst 1 x (hide x {x=y}) = {}
subst 1 x {x=y} = {y=1}
```

jonsterling: Oh sorry, I misread the code. Now I see, and I think your change is correct.

src/core/NewDomain.ml
```diff
@@ -900,8 +900,13 @@ struct
 
   let occurs xs =
```
jonsterling: oh lol, you were right! I forgot that I had written something like this :laughing:

src/core/NewDomain.ml
```diff
@@ -662,14 +724,14 @@ struct
 
     | Neu info ->
       let neu = {head = NCoe {r; r'; ty = Abs (x, info.neu); cap}; frames = Emp} in
-      let ty = Val.subst_then_run (Rel.hide' x rel) r x tyx in
+      let ty = Delayed.make' (Some (Rel.hide' x rel)) (subst r' x tyx) in
```
favonia: I think the previous version, which was trying to substitution `<r/x>`, is wrong. It should be `r'/x`.

favonia: Never mind, this was introduced by me in ead41d1a15030aa1ad3dda94d185a6f7204e10ab LOL.

src/core/NewDomain.ml
```diff
@@ -1275,7 +1276,12 @@ and Face :
       (r, r', Delayed.make @@ f (r, r', Delayed.drop_rel bdy))
   end
 
-and Abs : functor (X : Domain) -> Domain with type t = X.t abs =
+and Abs : functor (X : Domain) ->
+  sig
+    include Domain with type t = X.t abs
+    val inst : rel -> t -> dim -> X.t
```
favonia: @jonsterling New, simple API to remove misconception of `run` inside an abstraction.

jonsterling: :+1:

favonia:
This PR will close #360 along the way.

favonia:
Fixes #154 and fixes #360.

src/core/NewDomain.ml
```diff
@@ -0,0 +1,1898 @@
...
+  type t
+  val freshen_name : Name.t -> Name.t * t
+  val freshen_names : Name.t bwd -> Name.t bwd * t
+  val swap_name : t -> Name.t -> Name.t
+  val fold : (Name.t -> Name.t -> 'a -> 'a) -> t -> 'a -> 'a
+end =
+struct
+  type t = (Name.t * Name.t) list (* favonia: a demonstration of my laziness *)
+  let mimic x = Name.named (Name.name x)
+  let freshen_name x =
```
jonsterling: As a matter of style, I like to put newlines between functions, so that in vim I can navigate the code using `{` and `}`.

favonia: fixed (?)

favonia:
@jonsterling Any Thoughts on `NewTyping` before it's too late?

jonsterling:
@favonia do you have specific questions about `NewTyping`, or are you just hoping for general Thought?

My only Thought about it is that the old version was really error-prone because I lazily just copied and pasted the slight-different-but-mostly-the-same code for checking every kind of system, instead of trying to use some smarter decomposition.

src/core/NewDomain.ml
```diff
@@ -179,13 +179,13 @@ module type Domain =
 sig
   type t
 
-  (** [swap] is a purely syntactic operation which perserves the property of being a Ξ-value. *)
+  (** [swap] is a purely syntactic operation which preserves the property of being a [rel]-value. *)
```
favonia: @jonsterling Taste it!

src/core/NewDomain.ml
```diff
@@ -1164,22 +1170,15 @@ struct
...
         in
-        let old_faces =
-          ConSys.foreach_gen info.sys @@ fun (s, s', lazy ty) ->
-          let rel' = Rel.equate' s s' rel in
-          let ty = run rel' ty in
-          let cap = Val.run rel' cap in
-          let sys = ConAbsSys.run rel' sys in
-          make_hcom rel' r r' ~ty ~cap ~sys
-        in
+        let old_faces = ConSys.plug rel nhcom info.sys in
```
favonia: It seems the `old_faces` can just be generated by `plug`...?

src/core/NewRestriction.ml
```diff
@@ -154,13 +154,13 @@ let equate' x y h =
   get_m h @@ equate x y h
 
 let subst (r : dim) (x : atom) (h : t) =
-  if mem_atom x h.index then
+  if mem_atom x h.index && r <> `Atom x then
```
favonia: @jonsterling I was blessed by the God of Syntax today!

jonsterling: :cloud_with_lightning: :crown:

favonia:
About 3bcc876c20363dbb4c4d88bf8969858db1d7782c: it's possible to optimize `swap`, but the code wrongly assumed the refreshing is just for one variable. Therefore it was reverted. We can unleash the correct optimization later.

src/core/NewDomain.ml
```diff
@@ -363,8 +366,10 @@ struct
     | Tm.Next _ ->
       raise CanJonHelpMe
 
-    | Tm.Data _ ->
-      raise CanJonHelpMe
+    | Tm.Data info ->
+      let params = List.map (fun t -> Val (Delayed.make @@ lazy (eval rel env t))) info.params in
```
jonsterling: @favonia Is this the way i'm supposed to write this code? Or is something else preferred?

src/core/NewDomain.ml
```diff
@@ -1388,12 +1391,11 @@ struct
...
         function
         | `Const tv ->
-          (* Favonia: I just want to put tv.value into there, but I'm a little owned by the types. -- JS *)
-          [Val (raise CanFavoniaHelpMe)]
+          [Val (LazyVal.make_from_delayed tv.value)]
         | `Rec (`Self, v) ->
           let v_ih = Val.plug rel ~rigid:true frm v in
-          (* Favonia: I want to put [v; v_ih] below, but I'm again owned by the types. -- JS *)
-          [Val (raise CanFavoniaHelpMe); Val (raise CanFavoniaHelpMe)]
+          [Val (LazyVal.make_from_delayed v);
```
jonsterling: LOL Thank you :rofl:

src/core/NewDomain.ml
```diff
@@ -2910,6 +2912,9 @@ end
 
     let make_from_lazy (lazy v) = make v
 
+    let make_from_delayed : X.t abs Delayed.t -> DelayedX.t abs =
+      Delayed.fold @@ fun rel (Abs (x, c_x)) -> Abs (x, Delayed.make' (Option.map (Rel.hide' x) rel) c_x)
```
favonia: Maybe we should just force it? (This function is new so was never used.)

src/core/NewDomain.ml
favonia: :rofl:

src/core/NewDomain.ml
```diff
@@ -0,0 +1,3041 @@
...
+      Neu {ty = Val.make ty; neu = {neu; sys = neu_sys}}
+
+    | _ ->
+      raise PleaseRaiseProperError
+
+
+  and rigid_multi_coe r r' data_abs args =
+    let Abs (x, ty) = data_abs in
+    (* Is there a way to write this code where I don't need to look up the argument specs in the signature? I hope so...
+       Otherwise, I need to thrad the global environment through too much stuff.
```
jonsterling: I think what I should do is bung into `Data` a closure of the `constrs` together with an **global** environment which can be used to evaluate them. It won't be necessary for it to respond to `run` or substitution or anything like that, because it is global.

src/core/NewDomain.ml
```diff
@@ -1759,12 +1765,35 @@ struct
...
+            Syn.eval_tm_sys rel benv tboundary
+          in
+
+          let fix_face =
+            ConAbsFace.gen @@ fun s s' (el : con) ->
+            let rel_ss' = Rel.equate' s s' rel in
+            let elx = make_coe rel_ss' (`Atom x) r' ~abs:(ConAbs.run rel_ss' abs) @@ Val.make el in
+            (* TODO: Favonia, is this actually safe? Shouldn't I be freshening? Also I feel weird
+               about all this unleash/make.*)
+            Abs (x, elx)
```
jonsterling: @favonia I feel ultra-iffy about this?

favonia: The code is wrong. (Sorry.) But let me unleash something...

src/core/NewDomain.ml
```diff
@@ -1787,9 +1789,8 @@ struct
             ConAbsFace.gen @@ fun s s' (el : con) ->
             let rel_ss' = Rel.equate' s s' rel in
             let elx = make_coe rel_ss' (`Atom x) r' ~abs:(ConAbs.run rel_ss' abs) @@ Val.make el in
-            (* TODO: Favonia, is this actually safe? Shouldn't I be freshening? Also I feel weird
-               about all this unleash/make.*)
-            Abs (x, elx)
+            let x', pi = Perm.freshen_name x in
+            Abs (x', Con.swap pi elx)
```
favonia: This is unfortunately wrong because `el` might have access to the outside `x` which was then shadowed. I wanted to fix the code but am not sure which part can depend on the inner, bound `x`. Can `data` and `genv` depend on `x`, for example?

jonsterling: `data` can depend, but `genv` cannot.

favonia: So `constr`, `constrs`, `data`, `tboundary`, `env_ps`, `sys` and `intro` are all morally under the binder `x`, and `genv` refers to the global env so it cannot depend on `x` in any way? Did I miss something?

I am worried about the innermost `benv` which extends `env_ps`. Should we apply `subst r x` to `env_ps`? Anyways I am not sure which part is under the binder `x`. If something only makes sense under the `x` binder, `Rel.hide' x rel` needs to be used instead of `x`.

favonia: It will help me a lot if you could rename or mark the variables in some systematic way with the dependency and shadowing.

jonsterling: I'm a little puzzled by all the shadowing issues. Shouldn't we just be freshening when we unbind???

favonia: Maybe we should, and then that will be at right after `let Abs (x, data_tyx) = data_abs in`.

jonsterling: I feel that any time we pattern match on `Abs (x, t)`, we are probably introducing a bug. It's most likely impossible to perfect the invariants generally so that this is safe.

We should have some "unleash" function or something, like we used to, and use this all the time. Then, in specific cases where we know what we are doing *locally*, we can do something more efficient, like doing the unsafe match.

Does this make sense to you?

favonia: Mostly agreed, except that I want to give the carefully crafted V and fhcom types a try.

jonsterling: OK cool! Let's have a safe API available so that low-dimensions people like me can use it, but allow people who have grasped the Geometry to use `Abs` directly.

favonia:
- [ ] optimize NewRestriction with PersistentTable.size.

favonia:
- [ ] TODO: optimize NewRestriction with PersistentTable.size

src/core/NewRestriction.ml
```diff
@@ -0,0 +1,213 @@
...
+  let rst = equate' x `Dim0 @@ emp () in
+  assert (compare x `Dim0 rst == `Same)
+
+
+
+
+
+let from_old_restriction rst =
+  let eqns = Restriction.chronicle rst in
+  List.fold_left (fun rel (r, r') -> equate' r r' rel) (emp ()) eqns
```
favonia: Looks okay to me now. Hope this will be killed by the New Order eventually.

jonsterling: yes! Once we can delete all the old stuff, we can delete this too.

src/frontend/Importer.ml
```diff
@@ -26,20 +26,22 @@ struct
...
+      end
     end
 
   let run_and_rot ~mlconf ~mlcmd:{con; span} =
     isolate_module ~mlconf begin
-      try_ begin
+      begin
         Elab.eval_cmd con >> abort_unsolved span >> RotIO.write
-      end cool_error_printer
+      end
```
favonia: Is there a reason to turn off ^^cool^^ error printing?

jonsterling: I turned it off for the sake of debugging... it was otherwise too hard to find the source of some exception. But I didn't mean to commit this...

favonia:
Right now `(vin 0 tm garbage)` will probably type-check. Is it what we want? Same thing for boxes.

This means our type-checking algorithm is not strictly speaking sound with respect to the current rules on paper, but I believe it is still consistent.

jonsterling:
@favonia This is intentional; since the alternative is to add a mountain of annotations, I think this is the most practical option.

jonsterling:
(And it's easier to start from here and then add enough annotations, than to go the other way around.)

favonia:
A subtle, conceptual bug was found in the current design: the delayed restrictions in faces might become inconsistent due to substitutions. There are several ways to fix this problem:

0. Merge `subst` and `run` into one single function in the `Domain` interface, or
1. Use `FortyTwo` or drop restrictions in `Face.subst` when there is inconsistency, or
2. Put delayed restrictions in `Lazy.t`, or
3. Make all faces as closures.

Personally, I prefer 0 because it fixes this bug and eliminates the possibility of forgetting `run` after `subst`. Solution 1 is also good. Solution 3 is probably the best after this PR is done.

jonsterling:
@favonia is there a simple example of where the substitution makes some restriction inconsistent? 
(It's ok to say no :laughing:)

favonia:
> @favonia is there a simple example of where the substitution makes some restriction inconsistent?
> (It's ok to say no)

Before I updated the code, `ConSys.subst 1 i (ConSys.run empty [i=0 -> foo])` I think will trigger the bug.

favonia:
The current hack is based on solution 1b---dropping the restriction completely when it becomes inconsistent.

src/core/NewTyping.ml
```diff
@@ -1058,7 +1059,6 @@ and synth_stack cx vhd ty stk  =
       let frm = D.Elim {lbl = data.lbl; params = data.params; mot = mot_clo; clauses} in
       D.Val.plug (Cx.rel cx) frm vhd
     in
-    let ty = inst_clo cx mot_clo @@ D.Val.unleash vhd in
```
jonsterling: oops :laughing:

library/prelude/path.red
```diff
@@ -121,8 +121,6 @@ def J/eq
   | k=1 → refl
   ]
 
-quit
-
 def symmd (A : 𝕀 → type) (p : (i : 𝕀) → A i) : pathd (symm^1 _ A) (p 1) (p 0) =
```
jonsterling: @favonia This raises some stupid `Dead` exception.  I guess this is what you are already looking at?

jonsterling:
Making progress! :smile: Now there's some mortality raising Neutroid.Triv.

src/core/NewDomain.ml
```diff
@@ -1054,7 +1054,7 @@ struct
       Format.fprintf fmt "<con>"
 
   let make_arr rel ty0 ty1 =
-    let env = Env.init_isolated [Cell.value ty0; Cell.value ty1] in
+    let env = Env.init_isolated [Cell.value ty1; Cell.value ty0] in
```
favonia: :rofl:

favonia:
The current bug is that some system about V types sent to `check_of_ty` is not fully evaluated. If I added a supposedly redundant `run` in `check_boundary` then `univalence` would typecheck.

favonia:
Milestone: `univalence` type checks.

src/core/NewDomain.ml
jonsterling: Lmao wtf!!!

favonia:
The new functions revealed potent bugs even in `path.red`... :crying_cat_face:

favonia:
The new functions revealed latent or conceptual bugs even in `path.red`... :crying_cat_face:

favonia:
@jonsterling BTW we both put our own `pp_list` in different files :rofl:

favonia:
Now we are back to `/home/favonia/documents/records/code/redtt/library/paths/sigma.red`...

favonia:
Now we are back to `library/paths/sigma.red`...

src/core/NewDomain.ml
```diff
@@ -0,0 +1,3788 @@
...
+        let sys = Syn.eval_tm_sys rel benv tboundary in
+        make_intro rel ~dlbl:data.lbl ~clbl:intro.clbl ~args:(coerced_args r r') ~sys
+      in
+      begin
+        match tboundary with
+        | [] -> intro
+        | _ ->
+          let faces =
+            let args_rx = coerced_args r @@ `Atom x in
+            let benv
```
jonsterling: The OCaml compiler may be unhappy, and I am also unhappy with `exception Triv` :wink:  I bet we'll find a lot of bugs if we get rid of such things elsewhere too.

favonia:
@jonsterling we have made significant progress. I believe the quotation of `NCoeData` is the last component to type check `paths/s1.red`!

Issue #253, opened 2018-08-19
Codes, HITs, and composition


ecavallo:
```
import path
import unit

data O where
| obase
| oloop (y : O) @ i
  [ i=0 ⇒ y
  | i=1 ⇒ y
  ]

let Code (a : unit) : type =
  elim a [ triv ⇒ O ]

let encode (p : Path unit triv triv) : O =
  coe 0 1 obase in λ i → Code (p i)

let decode (o : O) : Path unit triv triv =
  elim o [
  | obase ⇒ λ _ → triv
  | oloop (o' ⇒ decode/o') i ⇒ λ j →
    comp 0 1 (decode/o' j) [
    | i=0 ⇒ λ _ → decode/o' j
    | i=1 ⇒ λ _ → decode/o' j
    | j=0 ⇒ λ _ → triv
    | j=1 ⇒ λ _ → triv
    ]
  ]

let test (o : O) (i : dim)
  : Path O (encode (decode (oloop o i))) (oloop o i)
  =
  ?
```
raises `Failure("equate_dim: dimensions did not match")`. The error disappears if I delete the boundary of `oloop`, replace it with something non-recursive, redefine `Code` as just `O`, or replace the composition in `decode` with just `decode/o' j`.

jonsterling:
We've been having the same error message in some examples from @mortberg, thanks for this further example! I have been having a lot of trouble tracking down the problem, but for the example that I had before, what appeared to be happening was that `quote_nf` was being called, and then somehow in one of the recursive calls to `equate`, the left and right hand sides ceased to be equal (even though the initial call is putting the *same* thing on the left and right). In particular, a dimension was getting fucked up.

jonsterling:
Here's an even simpler version:

```
data O where
| obase @ i []

let encode_ty (p : Path unit triv triv) (i : dim) : type =
  elim (p i) [ triv ⇒ O ]

let decode (o : O) : Path unit triv triv =
  elim o [
  | obase i ⇒ λ j →
    comp 0 1 triv [
    | j=0 ⇒ λ c → triv
    | j=1 ⇒ λ d → triv
    | i=0 ⇒ λ _ → ?
    ]
  ]

let bad-ty (o : O) (k : dim) : _ =
  coe 0 1 (obase 0) in encode_ty (decode (obase k))

normalize bad-ty

```

the last call fails

ecavallo:
Not sure if this is related, but
```
data O where
| obase @ i []

[...]

PR #275, opened 2018-09-01
fix a bug in coe of extension type; various other cleanup

jonsterling:
(Don't get your hopes up, this doesn't fix the bug we were trying to fix .but it's clearly wrong!)


jonsterling:
Fwiw, this does allow us to make a bit more progress on @mortberg's `problem` file. It does not help with Evan's example (#253).

PR #279, opened 2018-09-02
fix boundary fix

ecavallo:
The `fhcom` should go from `r'` to `r`. (Until we have parameterized HITs, this is very hard to trigger.)


jonsterling:
Thanks!!

PR #281, opened 2018-09-02
An attempt to have a better grammar. Close #158.

### LINKED ISSUE #158: Allow fancy syntax (m n o : nat) -> blah?

favonia:
Really wanted to change `ESig` but I can still project my thought.

@jonsterling I noticed that many files under `examples` are missing in the `Makefile` and thus not checked. Should we just check everything under `examples`?

### DIFF
--- a/example/pointed.red
+++ b/example/pointed.red
@@ -6,23 +6,23 @@ let ptype : type^1 = (A : type) × A
 let pmap (pA pB : ptype) : type =
   (f : pA.0 → pB.0) × Path _ (f pA.1) pB.1
 
-let p→ (pA pB : ptype) : ptype = < pmap pA pB, λ a → pB.1 , λ i → pB.1 >
+let p→ (pA pB : ptype) : ptype = ( pmap pA pB, λ a → pB.1 , λ i → pB.1 )
 
 let pequiv (pA pB : ptype) : type =
   (f : pmap pA pB) × IsEquiv pA.0 pB.0 f.0
 
-let pbool : ptype = < bool , ff >
+let pbool : ptype = ( bool , ff )
 
 let pf (pA : ptype) : pequiv (p→ pbool pA) pA =
   let fwd : pmap (p→ pbool pA) pA =
-    <λ f → f.0 tt , λ i → pA.1>
+    (λ f → f.0 tt , λ i → pA.1)
   in
   let bwd : pA.0 → (pmap pbool pA) = λ a →
-    < λ b → elim b [
+    ( λ b → elim b [
             | tt ⇒ a
             | ff ⇒ pA.1
             ]
-    , λ i → pA.1 >
+    , λ i → pA.1 )
   in
   let bwdfwd : (f : pmap pbool pA) → Path (pmap pbool pA) (bwd (fwd.0 f)) f =
     λ f →
@@ -42,8 +42,8 @@ let pf (pA : ptype) : pequiv (p→ pbool pA) pA =
           ]
       in
       λ i →
-        < λ b → bwdfwd/map b i
+        ( λ b → bwdfwd/map b i
         , λ j → bwdfwd/pt i j
-        >
+        )
   in
-  <fwd, (Iso/Equiv _ _ <fwd.0, bwd, λ a i → a, bwdfwd>).1>
+  (fwd, (Iso/Equiv _ _ (fwd.0, bwd, λ a i → a, bwdfwd)).1)
--- a/example/pointed.red
+++ b/example/pointed.red
@@ -3,13 +3,13 @@ import isotoequiv
 
 let ptype : type^1 = (A : type) × A
 
-let pmap (pA,pB : ptype) : type =
-  (f : (pA.0) → (pB.0)) × Path _ (f (pA.1)) (pB.1)
+let pmap (pA pB : ptype) : type =
+  (f : pA.0 → pB.0) × Path _ (f pA.1) pB.1
 
-let p→ (pA,pB : ptype) : ptype = < pmap pA pB, λ a → pB.1 , λ i → pB.1 >
+let p→ (pA pB : ptype) : ptype = < pmap pA pB, λ a → pB.1 , λ i → pB.1 >
 
-let pequiv (pA,pB : ptype) : type =
-  (f : pmap pA pB) × IsEquiv (pA.0) (pB.0) (f.0)
+let pequiv (pA pB : ptype) : type =
+  (f : pmap pA pB) × IsEquiv pA.0 pB.0 f.0
 
 let pbool : ptype = < bool , ff >
 
@@ -17,7 +17,7 @@ let pf (pA : ptype) : pequiv (p→ pbool pA) pA =
   let fwd : pmap (p→ pbool pA) pA =
     <λ f → f.0 tt , λ i → pA.1>
   in
-  let bwd : (pA.0) → (pmap pbool pA) = λ a →
+  let bwd : pA.0 → (pmap pbool pA) = λ a →
     < λ b → elim b [
             | tt ⇒ a
             | ff ⇒ pA.1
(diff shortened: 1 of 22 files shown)


src/frontend/Grammar.mly
```diff
@@ -71,45 +89,86 @@ debug_filter:
...
-      | 0 -> E.Car
-      | 1 -> E.Cdr
-      | _ -> failwith "Parser: invalid projection" }
+vertebrae:
+  | e = with_projs(atom_econ)
+    { e }
+  | e = with_projs(atomic_econ)
+    { e }
 
+app_spine_as_pair:
```
favonia: Any idea for a better name?

src/frontend/Grammar.mly
```diff
@@ -190,20 +242,20 @@ epatbind:
...
 
 
 escheme:
-  | tele = list(etele_cell); COLON; cod = eterm
+  | tele = list(etele_cell); COLON; cod = located(econ)
     { (List.flatten tele, cod) }
 
 etele_cell:
-  | LPR; xs = separated_nonempty_list(COMMA, ATOM); COLON; ty = eterm; RPR
+  | LPR; xs = nonempty_list(ATOM); COLON; ty = located(econ); RPR
```
favonia: TASTE IT!

favonia:
@cangiuli @ecavallo Now `a b .0` is parsed as `a (b.0)` not `(a b).0`. Any Thought?

example/pointed.red
```diff
@@ -26,15 +26,15 @@ let pf (pA : ptype) : pequiv (p→ pbool pA) pA =
...
         =
         λ i j →
-          comp 1 j (pA.1) [
+          comp 1 j pA.1 [
           | i=0 ⇒ refl
           | i=1 ⇒ f.1
           ]
       in
-      let bwdfwd/map : (b : bool) → Path (pA.0) ((bwd (fwd.0 f)).0 b) (f.0 b) =
+      let bwdfwd/map : (b : bool) → Path pA.0 ((bwd (fwd.0 f)).0 b) (f.0 b) =
```
favonia: The parentheses in the old style were not needed, actually.

example/univalence.red
```diff
@@ -249,7 +249,7 @@ let univalence/alt (B : type) : IsContr^1 ((A : type) × Equiv A B) =
            let ctr/B : dim → B =
              λ j →
                comp 1 j b [
-               | i=0 ⇒ w .1 .1 b .0 .1
+               | i=0 ⇒ (w.1.1 b).0.1
```
favonia: Another case.

example/univalence.red
```diff
@@ -261,9 +261,9 @@ let univalence/alt (B : type) : IsContr^1 ((A : type) × Equiv A B) =
                let filler : dim → B =
                  λ l →
                    comp 1 l b [
-                   | i=0 ⇒ w .1 .1 b .1 v j .1
-                   | i=1 ⇒ λ k → connection/or B (v .1) j k
-                   | j=0 ⇒ v .1
+                   | i=0 ⇒ ((w.1.1 b).1 v j).1
```
favonia: Yet another case.

example/univalence.red
```diff
@@ -107,11 +107,11 @@ let PropIsContr (A : type) : IsProp (IsContr A) =
...
-let PropIsEquiv (A,B : type) (f : A → B) : IsProp (IsEquiv A B f) =
+let PropIsEquiv (A B : type) (f : A → B) : IsProp (IsEquiv A B f) =
   λ e0 e1 i b → PropIsContr (Fiber A B f b) (e0 b) (e1 b) i
 
 ; A direct proof that IsEquiv f is a prop, ported from cubicaltt to yacctt to redtt
-let PropIsEquivDirect (A,B : type) (f : A → B) : IsProp (IsEquiv A B f) =
+let PropIsEquivDirect (A B : type) (f : A → B) : IsProp (IsEquiv A B f) =
   λ ise ise' i y →
     let a : A = (ise y).0.0 in
     let p : Path B (f a) y = (ise y).0.1 in
```
favonia: All of these parentheses were not needed but now they are required.

cangiuli:
I prefer these changes to the original behavior, but I'm not sure how everyone feels.

favonia:
It seems everyone likes the new behavior. Now I am only waiting for some approval of the PR.

favonia:
> @jonsterling I noticed that many files under examples are missing in the Makefile and thus not checked. Should we just check everything under examples?

Let me point out that this will result in some files being checked multiple times. On the other hand, we have been missing important files like `pointed.red`!

jonsterling:
Yes, I think so. Should we revoke approval of this PR pending its implementation then? Ugh

src/frontend/Grammar.mly
```diff
@@ -71,45 +95,87 @@ debug_filter:
...
+  | p = eproj
+    { p }
 
+framic:
+  | e = located(atomic)
+    { E.App e }
+  | p = eproj
+    { p }
+
+spine:
```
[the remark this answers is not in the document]
jonsterling: (BTW in case it wasn't clear, I'm not objecting to it! Just wanted to understand it)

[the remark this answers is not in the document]
favonia: It's trying to keep atoms as `nonempty_list(ATOM)` till the last symbol. The critical case is the `d` position in the following state:
```
(a b c d : e) -> f
(a b c d .0)
(a b c d ^1)
...
```
Menhir refuses to make any decision before the symbol following `d`. Therefore, the `App` frames can only be reconstructed after we are sure `a b c d` are spines, not bound variables.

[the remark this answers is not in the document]
favonia: The current version has the disadvantage that the individual location information of `b c d` in the above example is lost, because `nonempty_list(ATOM)` is not keeping any of it. The solution is to replace `nonempty_list(ATOM)` with something that remembers the location. @jonsterling Do you think it's important?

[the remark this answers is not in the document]
jonsterling: @favonia OK, cool! Thanks very much for figuring this out. It's exactly the case that prevented me from being able to make the grammar how I wanted it using just my Weak Thought :laughing: 

For now, I'm OK with disrupting those source locations for now, we can clean it up in another pass later on. I'd be happy to do so.

src/core/Tm.ml
```diff
@@ -786,9 +786,9 @@ and pp_cmd env fmt (hd, sp) =
     | Snoc (sp, f)->
       match f with
       | Car ->
-        Format.fprintf fmt "@[<hv1>(car@ %a)@]" (go `Car) sp
+        Format.fprintf fmt "@[<hv1>%a .fst@]" (go `Car) sp
```
jonsterling: I think this should be `(fst %a)`, since it's the core language. (Or else, we should change how the core language is parsed, which is another option.)

favonia: I thought about it, and realize we should probably do a more thorough change with s/Cdr/Fst/ or something... I can do the Lisp style for now.

jonsterling:
Ah, I didn't realize that this didn't contain the change `.0` to `.fst`. I can do that in a separate PR.

vim/syntax/redtt.vim
```diff
@@ -31,7 +31,7 @@ syn keyword redttKeyw open shut tick dim prev next dfix fix
 
 syn keyword redttDecl opaque let debug normalize import quit
 
-syn match   redttSymb '[#@`|\[\]^*×:,.∙✓□▷=⇒→<>λ]\|->'
+syn match   redttSymb '[#@`|\[\]^*×:,.∙✓□▷=⇒→()λ]\|->'
```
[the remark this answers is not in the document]
favonia: I do want some highlighting. Any Thought to highlight matched parentheses as well?

PR #286, opened 2018-09-02
Port problem.ctt

ecavallo:
From https://github.com/mortberg/cubicaltt/blob/pi4s3/examples/problem.ctt.

redtt gets stuck on `innerpath`. @mortberg

### DIFF
--- a/example/pointed.red
+++ b/example/pointed.red
@@ -47,3 +47,21 @@ let pf (pA : ptype) : pequiv (p→ pbool pA) pA =
         >
   in
   <fwd, (Iso/Equiv _ _ <fwd.0, bwd, λ a i → a, bwdfwd>).1>
+
+let pΩ (pA : ptype) : ptype =
+  < Path (pA.0) (pA.1) (pA.1)
+  , λ _ → pA.1
+  >
+
+let pΩ/map (pA, pB : ptype) (pf : pmap pA pB) : pmap (pΩ pA) (pΩ pB) =
+  < λ p i →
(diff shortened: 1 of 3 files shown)


example/pointed.red
```diff
@@ -46,4 +46,22 @@ let pf (pA : ptype) : pequiv (p→ pbool pA) pA =
         , λ j → bwdfwd/pt i j
         )
   in
-  (fwd, Iso/Equiv _ _ (fwd.0, bwd, λ a i → a, bwdfwd) .1)
+  (fwd, (Iso/Equiv _ _ (fwd.0, bwd, λ a i → a, bwdfwd).1))
```
favonia: There are some extra parentheses that I do not understand.

example/problem.red
```diff
@@ -0,0 +1,38 @@
...
+  )
+
+let ps2 : ptype = (s2, base)
+let ps3 : ptype = (s3, base)
+let pjoin : ptype = (join, inl base)
+
+let test0-2 : pΩ³ ps3 .0 =
+  λ i j k → cube i j k
+
+let f3 : (pΩ³ ps3 .0) → (pΩ³ pjoin .0) =
```
favonia: You should not need these parentheses around the types unless you really want to. (I do not object to adding them, but want to know whether there are still bugs in the parser.)

example/pointed.red
jonsterling: and let's put the space back before `.1`

ecavallo:
OK, I guess we should merge this for now.

PR #287, opened 2018-09-02
Simplify vproj

jonsterling:
Now, `vproj` needs only the data that appears in Part III. Yay!


mortberg:
Awesome! Funcy!

jonsterling:
@mortberg 
thanks! I'm surprised it worked

PR #293, opened 2018-09-05
fix s3-to-join, maybe

ecavallo:
The current definition of this "equivalence" is actually homotopic to a constant map. This one is less likely to be, since none of the boundary-fixing uses of `push` involve all the variables `i`,`j`,`k`. But we should really prove it is an equivalence to be sure.

@mortberg I bet this version would look real nice in cubicaltt

### DIFF
--- a/example/s3-to-join.red
+++ b/example/s3-to-join.red
@@ -1,6 +1,7 @@
 import path
 import s1
 import s2
+import isotoequiv
 
 data s3 where
 | base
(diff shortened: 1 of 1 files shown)


favonia:
I wonder what happened to travis...

mortberg:
I ported this to cubicaltt now: https://github.com/mortberg/cubicaltt/commit/a91feccdd5e7db7228309470ff1057953127ebba

I'ts slightly more complicated than the old wrong version, but it's not too bad

favonia:
The new version is even better (in Cartesian settings).

favonia:
This is amazing, and we should merge this PR immediately once the equivalence is proved unless @ecavallo says otherwise.

PR #295, opened 2018-09-06
Cleanup

jonsterling:
Using this PR to clean up a bunch of cruft that is bogging me down. Stay tuned.

### DIFF
--- a/example/Makefile
+++ b/example/Makefile
@@ -5,4 +5,4 @@ DUNE=${EXEC} dune --
 RED_FILES=$(wildcard *.red)
 
 all:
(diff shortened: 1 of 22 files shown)


jonsterling:
@favonia Maybe related in some complicated way, not sure... very strange!

jonsterling:
This makes name resolution less stupid, and also gives some better error messages (you used to see a lot of things like "can't find datatype named foo" when that wasn't at all the problem).

jonsterling:
There's more to be cleaned up, but let's get this one merged in first.

PR #296, opened 2018-09-07
Parser hacks.

favonia:


jonsterling:
Thanks for making the concrete notation less brutal :heart:

PR #297, opened 2018-09-07
apply Guillaume's Thought

ecavallo:
to define the Hopf fibration without using the expensive proof that UA IdEquiv is a constant path. With `coe` in datatypes redefined as identity, `innerpath` now checks instantaneously! (Without, it takes ~2sec.)

Waiting to see if `problem` gets anywhere...


mortberg:
Nice! It's funny how you can mimic a Glue type with idEquiv on the boundary using a comp with V-types with idEquiv's

ecavallo:
No luck with `problem` :(

PR #298, opened 2018-09-07
Tweak s3-join-s3.

favonia:
We are killing dead faces now.


cangiuli:
False Face Transform (FFT)

PR #299, opened 2018-09-07
More Cleanup

jonsterling:


jonsterling:
(This PR is a slight performance improvement across the board)

jonsterling:
@favonia The basic summary is that there was some kind of gnarly stuff that is hard to understand, and in order to replace it and have the sensible version be fast enough, I needed to have a fast way to determine if a term is locally-closed.

PR #300, opened 2018-09-07
Know your signatures

favonia:


src/frontend/Parse.ml
```diff
@@ -71,10 +71,10 @@ let loop lexbuf tokens =
       | I.HandlingError env ->
         begin match I.top env with
           | None ->
-            Lwt.return []
+            Lwt.fail_with "[parser] parsing failed"
```
favonia: I want to make it more specific, but am not sure what the correct message is.

PR #301, opened 2018-09-08
Even more cleanup; fancier let expressions

jonsterling:


jonsterling:
@favonia @ecavallo I decided to add fancier let-expressions which let you write `let foo (x : A) (y : dim) ... : B = ... in ...`.

PR #304, opened 2018-09-08
Loop space of S2

ecavallo:
This is not related to the Brunerie number, but it's a pet theorem I've tried to formalize a few times over the years. I got it from a [blog post of Egbert's](https://egbertrijke.com/2016/09/28/the-loop-space-of-the-2-sphere/), and I believe it's an instance of a more general theorem in his thesis (?).

The file takes a little under a minute to check on my machine. I've replaced the current `coe` in datatypes with the identity function in this branch, which is valid for datatypes without parameters. Without that change, it seems to get intractable around `s2/encode-decode/base/step`.

First `redtt`-exclusive formalized result (as far as I know) 🎉

### DIFF
--- a/example/omega1s2.red
+++ b/example/omega1s2.red
@@ -57,7 +57,7 @@ let onegloop-oloop (o : os2)
 
 ; II. universal cover over s2
 
-let S2CodeSurf/filler : (m i j : dim) → type =
+let s2/Code/Surf/filler : (m i j : dim) → type =
   λ m i j →
   comp 0 m os2 [
   | i=0 → UA os2 os2 (IdEquiv/wc os2)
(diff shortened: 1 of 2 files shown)


jonsterling:
Wow, cool!!!

example/omega1s2.red
```diff
@@ -0,0 +1,167 @@
...
+  | i=1 → refl
+  | k=0 → refl
+  | k=1 j → surf i j
+  ]
+
+let s2/decode/base (o : os2) : Path s2 base base =
+  elim o [
+  | obase → λ _ → base
+  | oloop (o' → s2/decode/base/o') i → λ k →
+      extend-by-surf s2/decode/base/o' i 1 k
```
jonsterling: I think here is an opportunity for an eta-reduction

example/omega1s2.red
```diff
@@ -0,0 +1,167 @@
...
+  ]
+
+let s2/decode (a : s2) : (S2Code a) → Path s2 base a =
+  elim a [
+  | base → λ o → s2/decode/base o
```
jonsterling: another eta reduction (hopefully)

PR #306, opened 2018-09-09
add notation for boundary of a set of dimensions??

jonsterling:
While we're golfing code, what do folks think about this notation?

/cc @mortberg @cangiuli @ecavallo @favonia

### DIFF
--- a/vim/syntax/redtt.vim
+++ b/vim/syntax/redtt.vim
@@ -22,7 +22,7 @@ syn region  redttEncl transparent start="\[" end="\]" contains=ALLBUT,redttBrack
 
 syn match   redttHole '?\k*'
 
-syn keyword redttKeyw boundary data intro where in with end elim
+syn keyword redttKeyw data intro where in with end elim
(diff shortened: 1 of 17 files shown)


favonia:
I wish there was an ASCII version, but could not find a good symbol.

jonsterling:
@favonia Yeah, I too was trying to figure out what would be a reasonable ascii version... :frowning:

vim/syntax/redtt.vim
```diff
@@ -22,7 +22,7 @@ syn region  redttEncl transparent start="\[" end="\]" contains=ALLBUT,redttBrack
 
 syn match   redttHole '?\k*'
 
-syn keyword redttKeyw data intro where in with end elim
+syn keyword redttKeyw boundary data intro where in with end elim
```
jonsterling: is this intended?

favonia: No. LOL

favonia:
@jonsterling I suppose we are both okay with the new syntax (?).

PR #310, opened 2018-09-11
Issue 309

### LINKED ISSUE #309: elim boundary calculated incorrectly
```
data void where

data cool where
| one [v : void] 
| two [v : void] (a : cool) @ i [ i=0 → one v | i=1 → a ]

let test (a : cool) : void =
  elim a [
  | one v → v
  | two v (a' → v') i → ?
  ]
```
results in a hole with the restrictions `i = 0 ⇒ a'` and `i = 1 ⇒ v'`, but the `i = 0` face should actually be `v`.

jonsterling:
This appears to resolve #309, but the code is very bad and needs to be both checked & cleaned up. @ecavallo Want to have a look?

### DIFF
--- a/src/core/Tm.ml
+++ b/src/core/Tm.ml
@@ -202,8 +202,22 @@ and face_info : type x. (x -> Info.t) -> (tm, x) face -> Info.t =
 and tm_info (Tm node) =
   node.info
 
+
 let make con =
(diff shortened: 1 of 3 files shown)


jonsterling:
I'll merge this. It's still ugly, but it's not as ugly as it was.

PR #311, opened 2018-09-11
Update syntax highlighter.

cangiuli:
* Syntax-highlight parentheses and brackets properly (@favonia)
* Update list of tokens from the parser.
* More aggressively look for matching parentheses (@jonsterling)


vim/syntax/redtt.vim
```diff
@@ -13,25 +10,23 @@ endif
 setlocal iskeyword=a-z,A-Z,48-57,-,',/
 
 syn sync minlines=50
+syn sync maxlines=1000
```
jonsterling: nice.

PR #312, opened 2018-09-11
first cut at imposing some lexical conventions; new comment style

jonsterling:
Comments and suggestions welcome

### DIFF
--- a/example/J.red
+++ b/example/J.red
@@ -2,8 +2,8 @@ import path
 
 let J
   (A : type) (a : A)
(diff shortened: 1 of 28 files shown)


jonsterling:
I have updated the comment notation to the new proposal. I also updated the Vim highlighter.

The one thing I couldn't figure out how to do was support nested comment blocks in Vim, which the redtt lexer does support. It's not a big deal, just thought I'd mention it.

#311 should be merged before this PR.

PR #318, opened 2018-09-14
update versions of compiler and dependencies

jonsterling:


jonsterling:
To upgrade your environment, just follow the commands in the README, including opam  switch.

PR #320, opened 2018-09-14
Try to use ordinary terms instead of formal boundary terms

jonsterling:


jonsterling:
This is an intermediate step between applying the @Favonia Rectification Campaign. To summarize:

- I had implemented a version of cubical inductive types that was just a restriction of what appears in Part IV. This includes the explicit use of formal boundary terms, as well as having a rigid kind of constructor schema, in which constant arguments and recursive arguments and dimension arguments are separated into groups.

- The lesson I learned is that, for implementation at least, this style makes things very hard. I need to write more complicated code, again and again (but slightly different each time). A more systematic way to implement this schema was pioneered by @favonia in RedPRL, which we will adopt, in which we do not require the arguments to a constructor to appear in a specific order (but typing ensures that they are sensible). The code is more systematic that way.

- As an intermediate step that will make the change easier, I have removed the explicit formal boundary terms and am just re-using the term language of redtt, but checking that the terms fall into the right fragment. This simplifies the code quite a bit; this was not done in the RedPRL implementation, but I got the impression that we would have liked to do so.

In order to make this work, I added a notion of "partial datatype declaration" to the signature, which means that a datatype is available with some constructors, but that you aren't yet allowed to call its eliminator (because more constructors will be added). As the datatype scheme is elaborated, it is constantly updated in the signature until it is marked `Complete`, at which point the eliminator becomes available.

@favonia @ecavallo

jonsterling:
This now has the side effect that you can do things like put holes inside a datatype declaration!

PR #322, opened 2018-09-15
Flexible HIT Spec: Stage 2

### LINKED ISSUE #208: Move dimension specs to the beginning of a constructor spec
This is an extension to Evan & Bob's schema suggested by @favonia

jonsterling:
(after #321)

Resolves #208

### DIFF
--- a/example/int.red
+++ b/example/int.red
@@ -6,8 +6,8 @@ import equivalence
 import isotoequiv
 
 data int where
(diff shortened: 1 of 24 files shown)


jonsterling:
This is getting close to the immortal @favonia-inspired version. I think I just need to rewrite the eliminator tactic now.

PR #329, opened 2018-09-17
More optimization with relaxed validity

favonia:
The most relaxed local validity---non-bipartiteness. (See #330 for the non-local validity.)

### DIFF
--- a/src/core/Typing.ml
+++ b/src/core/Typing.ml
@@ -136,13 +136,13 @@ let check_valid_cofibration ?xs:(xs = None) cofib =
           | `Atom x, `Dim0 | `Dim0, `Atom x ->
             if Hashtbl.mem ones x then true else
               begin
-                Hashtbl.add zeros x ();
+                Hashtbl.replace zeros x ();
                 go eqns
               end
           | `Atom x, `Dim1 | `Dim1, `Atom x ->
             if Hashtbl.mem zeros x then true else
               begin
-                Hashtbl.add ones x ();
+                Hashtbl.replace ones x ();
                 go eqns
               end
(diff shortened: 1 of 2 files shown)


src/core/Typing.ml
```diff
@@ -117,43 +117,58 @@ let check_dim_scope oxs r =
     | _ -> ()
 
 let check_valid_cofibration ?xs:(xs = None) cofib =
```
jonsterling: Is this the algorithm to check that bipartiteness condition? If so, could you mention that in the code?

favonia: Does it look better now?

src/core/Val.ml
```diff
@@ -906,7 +887,7 @@ struct
             Abs.make1 @@ fun w -> ext_apply (cdr (fixer_fiber phi)) [`Atom w]
           in
           let el1 = make_hcom (Dir.make `Dim1 r') info.ty1 (base I.idn r r') @@
-            force_abs_sys [face0; face1; face_diag; face_front]
+            force_abs_sys [face0; face_diag; face_front]
```
jonsterling: So, let me make sure I understand. The `face1` was `r=1`, making the entire old cofibration `r=0 \/ r=1 \/ r=r' \/ r'=0`. Now the cofibration is `r=0 \/ r=r' \/ r' = 0`; and this becomes valid under the new order because we can also take into account the direction `1~>r'`?

favonia: It is valid for one of the two reasons: Including the direction makes it valid. Bipartiteness checking also makes it valid.

PR #340, opened 2018-09-18
Too mortal

jonsterling:


src/core/Val.ml
```diff
@@ -1907,7 +1907,7 @@ struct
           let body =
             lazy begin
               let func = Value.act phi func in
-              apply func el
+              apply func @@ Value.act phi el
```
jonsterling: @favonia This seemed like a bug, but not sure. what do you think?

favonia: Indeed! However, is there a reason not to use the helper functions in Face.ml?

jonsterling: The main reason is I didn't understand how to use those functions.

jonsterling: Let me update it

src/core/Quote.ml
```diff
@@ -397,7 +397,7 @@ struct
       let tr, tr' = equate_dir env info0.dir info1.dir in
       let _ = equate_ty env info0.univ info1.univ in
       let ty = equate_neu env info0.ty info1.ty in
-      let ty_val = make @@ Up {ty = info0.univ; neu = info0.ty; sys = []} in
+      let ty_val = reflect info0.univ info0.ty @@ ValSys.from_rigid info0.ty_sys in
```
jonsterling: Taste it. This was the problem.

PR #347, opened 2018-09-21
Parametric hits (Third time's the charm)

jonsterling:


src/core/Typing.ml
```diff
@@ -416,11 +417,11 @@ and check_intro cx dlbl params constr tms =
...
 
   let _ : D.env =
-    List.fold_right2
-      (fun (_, spec) tm tyenv -> check_argument tyenv spec tm)
+    List.fold_left2
+      (fun tyenv (lbl, spec) tm -> check_argument tyenv lbl spec tm)
+      (D.Env.append V.empty_env params)
       (Desc.Constr.specs constr)
       tms
-      (D.Env.append V.empty_env params)
```
jonsterling: BTW my perfectly symmetric mind is owning me again; I am confused why I needed a fold_left? Did I reverse something else? (Please note, I never used `List.rev`)

jonsterling:
Here's a bug: 

If you try to write a recursive constructor in a parametric HIT, like `foo (x : T A)`, it will interpret this as a constant argument using the partially defined `T` type. What should really happen is:

1. This should be an error: I need to check what's going on, but somehow it's not properly checking that the type is real
2. The right way to write this anyway is `foo (x : T)`, even though `T` has parameters. Because this is just notation for defining a fixed point in a particular slice. (BTW, if you write this, then it does get read as a recursive argument).

jonsterling:
This PR is now "implemented", but it remains to debug it (by trying increasingly difficult examples).

jonsterling:
Here's an example that doesn't yet work (the call to the elim tactic at the bottom throws an exception):

```redtt
data F (A : type) where
| η (a : A)
| ☆ (s t : F)
| ε
| ass (s t u : F) (i : dim) [
  | i=0 → ☆ s (☆ t u)
  | i=1 → ☆ (☆ s t) u
  ]
| idn/r (s : F) (i : dim) [
  | i=0 → ☆ s ε
  | i=1 → s
  ]
| idn/l (s : F) (i : dim) [
  | i=0 → ☆ ε s
  | i=1 → s
  ]


data list (A : type) where
| nil
| cons (x : A) (xs : list)


let eval (A : type) : F A → list A =
  elim [
  | η a → ?
  | ☆ s t → ?
  | ε → ?
  | ass s t u i → ?
  | idn/r s i → ?
  | idn/l s i → ?
  ]
```

For what it's worth, all the path constructors are equally problematic in this example.

ecavallo:
Here's some bug:
```
data fun-s1 (A : type) where
| base
| loop (j : dim) [∂[j] → base]

let test (A : type) (j : dim) : fun-s1 A =
  coe 0 1 (loop j) in λ _ → fun-s1 A

normalize test
```

jonsterling:
good catch! My first guess is that in some other place I have failed to properly populate the environment, possibly with the parameters.

jonsterling:
@ecavallo fixed, thanks for catching!!

PR #369, opened 2018-10-07
less confusing notation for datatype parameters

### LINKED ISSUE #368: Proposal for better HIT declaration notation
Parametric HITs are written as follows:

```
data F (x : A) where
| con (y : B) (r : F) ...
...
```

This notation is not great for a couple reasons:

1. Conceptually it makes sense that the recursive arguments are called `foo` only and not `foo A`, but it may be confusing to a user who doesn't have in mind a schema, but is just thinking about programming.

2. How will we distinguish *indices* from parameters? There's some options, but all of them are kind of confusing.

I have an idea that should resolve both issues at once. Below is my proposed notation for parametric & indexed HITs:

```
(x : A) ⊢ data F (ix : I) ... where
| con (y : B) (r : F i) ...
...
```

We would unleash the fragment of this notation for parametric HITs first. The only problem is that there is no vim digraph for the turnstile, but I bet it is possible to add one in the vim mode (@cangiuli?). In fact, we should actually consider making the vim mode auto-replace various symbols with their immortal/unicode equivalents as you type, so users don't have to learn digraphs.

I have a related proposal for a very immortal notation to define functions by case/elimination trees in the style of Epigram.

/cc @favonia

jonsterling:
resolves #368

had to compromise a little bit to avoid grammar conflicts

### DIFF
--- a/vim/ftplugin/redtt.vim
+++ b/vim/ftplugin/redtt.vim
@@ -86,3 +86,5 @@ function! s:CloseBuffer()
 endfunction
 
 let b:did_ftplugin = 1
+
+dig !- 8866
--- a/vim/ftplugin/redtt.vim
+++ b/vim/ftplugin/redtt.vim
@@ -87,4 +87,4 @@ endfunction
 
 let b:did_ftplugin = 1
 
-dig !- 8866
+digraph !- 8866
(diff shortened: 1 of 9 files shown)


vim/ftplugin/redtt.vim
```diff
@@ -87,4 +87,4 @@ endfunction
 
 let b:did_ftplugin = 1
 
-dig !- 8866
+digraph !- 8866
```
cangiuli: Thank you, LOL.

jonsterling: :sunglasses:

PR #371, opened 2018-10-07
RedML, Phase I

jonsterling:


jonsterling:
What remains is to restore the printing of `Defined foo` going by.

jonsterling:
Alright! I think it's ready for merging. I'd like to stop working on this for now and resume fixing those CJHMs in `new-domain`. /cc @favonia

PR #372, opened 2018-10-08
Close #370.

### LINKED ISSUE #370: Installation script for vim mode
I sat down with a new user last week and we had to install the vim mode, and it was so hard to walk them through doing it! First, to explain that weird `~/.vim/pack/foo/start/vim-redtt` shit, and then, we ran the `cp` command but accidentally copied redtt's `vim` directory into that location instead of its contents... The whole thing is so frustrating.

I insist that we must have some script that an average user can just run from inside the `redtt` directory which installs the vim mode; this script should do the following things:

1. If there doesn't exist any `~/.vim/pack/???/start` folder, we create one; if there is such a folder already, let's try to use it.

2. Copy the appropriate files into there. If there is already some `vim-redtt` thing, we must overwrite it.

I am *not* concerned with edge cases, like if the user has two different `.../pack/???/...` folders; this will probably never happen, and if it does, it's the user's problem. Furthermore, we will assume that users will put the `redtt` binary in their path, so there is no need to do any configuration of `g:redtt_path` or whatever.

cangiuli:
Write a tiny installation script, and improve vim mode's documentation while I'm at it. If we like this, I'll update the RedPRL mode similarly.

### DIFF
--- a/vim/ftplugin/redtt.vim
+++ b/vim/ftplugin/redtt.vim
@@ -1,7 +1,7 @@
 " vim-redtt ftplugin
 " Language:     redtt
 " Author:       Carlo Angiuli
-" Last Change:  2018 August 13
+" Last Change:  2018 October 8
 
 if (exists("b:did_ftplugin") || !has('job'))
   finish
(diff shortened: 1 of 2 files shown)


vim/ftplugin/redtt.vim
```diff
@@ -79,12 +81,9 @@ function! s:EditWidth()
 endfunction
 
 function! s:CloseBuffer()
-  cclose
```
jonsterling: what's this do?

cangiuli: It closes the quickfix window, which is not currently opened by the redtt mode!

jonsterling: ah! cool.

jonsterling:
Beautiful! It works great. Thank you so much!

PR #374, opened 2018-10-09
[vim] turn the install script into a script

jonsterling:
The reason is, this will definitively prevent people from running this command from the wrong directory. So far everyone who has ever used the instructions (including me!) has first done so in the wrong directory, and copied a bunch of junk into their folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,12 +32,18 @@ and [cubicaltt](https://github.com/mortberg/cubicaltt).
 
 - experimental support for Fitch-style modal guarded recursion: ▷
 
+- RedML, a (very) rudimentary tactic language
+
 
 Features we intend to add in the near future:
 
-- user-extensible tactic language
-
 - namespacing
+
+- indexed higher inductive types
+
+- a type system for RedML
+
+- algebraic effects and handlers for RedML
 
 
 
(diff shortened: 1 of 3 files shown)


jonsterling:
@cangiuli I know you didn't want to do this, but my experience suggests that it may be the best way to prevent mistakes caused by not reading the instructions.

cangiuli:
OK, that's fair. Do you need to `+x` the script?

jonsterling:
@cangiuli I think I did, but I guess that didn't carry over into Git. any suggestions?

jonsterling:
@cangiuli Now that's what I call a user study!! :sunglasses:

jonsterling:
@cangiuli LOL, the reason why using that command you suggested had no effect is that it was already `+x` in git.

cangiuli:
I pulled and it created the file `+x`. Nice!

cangiuli:
By the way, you know that people might still run `vim/install.sh` from the main directory instead, right...?

jonsterling:
@cangiuli Yes, that's possible LOL. But I there is a much lower risk of this being a problem, because if they just copy and paste the command in the README, it will only succeed if they are in the right directory. Whereas before, if they copied and pasted the command, it would succeed but do something very bad, if they are in the wrong directory.

PR #376, opened 2018-10-10
Complete induction

kaonn:


cangiuli:
OK I guess I did make two non-cosmetic changes: deleted `import bool` and deleted the `inc` function.

library/induction.red
```diff
@@ -1,85 +1,81 @@
...
-		 λ n →  P'n n n (eq/implies/le n)
+def complete/implies/weak (P : nat → type) : complete/induction P → weak/induction P =
+  λ complete p0 ps → complete p0 (λ n f → ps n (f n (eq/implies/le n)))
 
+def weak/implies/complete : (P : nat → type) →
+  ((P' : nat → type) → weak/induction P') → complete/induction P =
+  λ P weak p0 ps →
+    let P' : nat → type = λ n → (k : nat) → (le k n) → P k in
+    let P'0 : P' zero = λ k k/le/0 → transport nat zero k (le/zero/implies/zero k k/le/0) P p0 in
+    let f : (n : nat) → (P' n) → (P' (suc n)) =
```
jonsterling: A slicker way to write this is to put `let f (n : nat) (p'n : P' n) : P' (suc n) = \ k ...`; this way we don't have to have as many lambdas.

library/induction.red
```diff
@@ -1,85 +1,81 @@
...
-		| suc n' →  \ p →  elim p []
-	]
+def le/zero/implies/zero (n : nat) : (le n zero) → path nat zero n =
+  elim n [
+  | zero → λ _ → refl
+  | suc n' → λ p → elim p []
+  ]
 
-def inc : nat → nat = (\ i →  suc i)
+def le/case (m : nat) : (n : nat) → (le n (suc m)) → or (path nat n (suc m)) (le n m) =
```
jonsterling: A slicker way to write this is to put `(m : nat)` after the colon, and then change the `elim m` to `elim`.

library/induction.red
```diff
@@ -1,85 +1,81 @@
...
+    ]
+  ]
 
-def le/suc (n: nat) : (m : nat) →  le n m → le (suc n) (suc m) =
-	elim n [
-	| zero →  \ _ _  →  triv
-	| suc n' →  \ m' l → l 
-	]
+def le/suc (n: nat) : (m : nat) → le n m → le (suc n) (suc m) =
+  elim n [
```
jonsterling: A slicker way to write this it to put `(n : nat)` to the right of the colon, and change `elim n` to `elim`.

library/induction.red
```diff
@@ -0,0 +1,85 @@
+import bool
+import nat
+import path
+import unit
+
+def transport (A : type) (M N : A) (p : path A M N) (C : A → type) : C M → C N = 
+	λ cm → coe 0 1 cm in λ i →  C (p i)
```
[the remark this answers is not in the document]
cangiuli: I agree with @jonsterling; just inline `transport`. It will make `weak/implies/complete` look much nicer.

cangiuli:
@kaonn I'm not going to make further changes right now; I just was tidying things up as I read the code, and inadvertently addressed a few (but not all) of @jonsterling's suggestions. Looks good!

Issue #198, opened 2018-08-06
Hierarchical names


jonsterling:
Lean has a really nice feature where names are hierarchical, so a single declaration `foo` might elaborate such that `foo` has a meaning, but also `foo.x` and `foo.y` etc. have meanings.

PR #380, opened 2018-10-11
Rectification of imports.

favonia:
This PR will support some rudimentary notion of "redlib".


jonsterling:
@cangiuli Soon we can start to rectify the structure of the library :smile:

cangiuli:
Should we be thinking about #198 while we're hammering out the details of library structure?

jonsterling:
@cangiuli For now, I want to think of that orthogonally. But when we implement it, it will likely become related. Unless I misunderstood what you had in mind.

src/basis/SysUtil.ml
```diff
@@ -0,0 +1,38 @@
...
+
+  (** portable and reasonable normalization of a mix of absolute and relative paths
+      into an absolute path based on chdir/getcwd. *)
+  val normalize_concat : dirs:string list -> rel_path:string -> string
+
+  (** run a {b gyve} and then restore the current cwd. *)
+  val protect_cwd : (string -> 'a) -> 'a
+end =
+struct
+  let module_to_rel_path ?extension accessors =
```
jonsterling: whoa! I didn't realize the `?` could be used like this. cool.

src/basis/SysUtil.ml
```diff
@@ -0,0 +1,38 @@
...
+sig
+  (** turn [["a"; "b"; "c"]] into ["a/b/c"]. note that [extension]
+      should {b not} include the period. *)
+  val module_to_rel_path : ?extension:string -> string list -> string
+
+  (** portable and reasonable normalization of a mix of absolute and relative paths
+      into an absolute path based on chdir/getcwd. *)
+  val normalize_concat : dirs:string list -> rel_path:string -> string
+
+  (** run a {b gyve} and then restore the current cwd. *)
```
jonsterling: lmao

src/frontend/Contextual.ml
```diff
@@ -118,45 +119,62 @@ let modify_mlenv f =
 let get_mlenv = get <<@> fun st -> st.mlenv
 
 
+let assert_top_level =
```
jonsterling: Hahaha, very good :+1:

src/frontend/ML.ml
```diff
@@ -5,7 +5,8 @@ type 'a info =
   {con : 'a;
    span : Log.location}
 
-module Env = PersistentTable.M
+module T = PersistentTable.M
+type mlconf = {base_dir : string; indent : string}
```
jonsterling: what is the `indent` in here do?

jonsterling: Now I understand!! 

My suggestion is to change this to `import_depth : int`, and generate the appropriate indent-string when we're printing.

favonia: @jonsterling I just thought it would be stupid to regenerate the string all the time. What would be the immortal API?

jonsterling: @favonia I think it's ok to regenerate the string at all times, for the sake of modularity.

favonia:
I suggest unleashing this PR before I Rectify the indentation mechanism unless there's some serious bug. @jonsterling (0) data types are subtle because of constructors, and (1) please check whether all the added visibilities in `Unify.ml`, especially ``Private``, are correct.

PR #384, opened 2018-10-11
Quotients are weakly effective

jonsterling:


library/complete-induction.red
```diff
@@ -49,7 +46,7 @@ def le/case : (m n : nat) → (le n (suc m)) → or (path nat n (suc m)) (le n m
     | zero → λ _ → inr triv
     | suc n' → λ p →
       elim (c n' p) [
-      | inl p → inl (λ i -> suc (p i))
+      | inl p → inl (λ i → suc (p i))
```
favonia: Can we have redlint to detect these? LOL

jonsterling: :laughing:

PR #391, opened 2018-10-12
Who cares about being eta-long?

favonia:
Eta-long? Who cares? :rofl: Maybe this will work?


jonsterling:
@favonia Have you tried informally comparing performance on some examples? I'm curious. I would imagine that this would have better performance than the old code.

jonsterling:
To clarify, this PR respects eta equivalence, but generates normal forms which might not be eta long. It is slightly faster.

PR #399, opened 2018-10-15
library Rectification Campaign

jonsterling:


src/frontend/Lex.mll
```diff
@@ -238,3 +241,15 @@ and read_string buf =
...
+    { cells := Snoc (!cells, lexeme lexbuf);
+      read_import cells lexbuf }
+  | "."
+    { read_import cells lexbuf }
+  | line_ending
+    { new_line lexbuf;
+      IMPORT (Bwd.to_list !cells) }
+  | eof
+    { IMPORT (Bwd.to_list !cells) }
+  | _ { failwith @@ "Invalid path component character: " ^ lexeme lexbuf }
```
favonia: did you consider
```
import x -- this is a very cool library
```

jonsterling: @favonia LOL good point!

jonsterling:
The first pass at the rectification campaign is complete (but we might make further changes for aesthetic reasons).

One subtlety is that not all files are being checked by Travis anymore, because of the folder structure. @favonia Do you have any thoughts on this? One possibility is to have some `all.red` file, but then we would need a way to ensure that it stayed in sync. Two possibilities for this appeared to us:

1. Allow `import foo.*`, and then have `all.red` just be hardcoded with the main directories of the library.
2. Have `all.red` auto-generated by the makefile somehow (I really don't favor this idea!!).
3. Alternatively I guess we could have the script walk through the folders, but this seems like a waste of time since things would get elaborated a million times for no reason.

src/frontend/Lex.mll
jonsterling: Fixed it

library/Makefile
```diff
@@ -2,7 +2,7 @@ OPAM=opam
 EXEC=${OPAM} config exec
 DUNE=${EXEC} dune --
 
-RED_FILES=$(wildcard *.red)
+RED_FILES=$(wildcard ./**/*.red)
```
jonsterling: @cangiuli @favonia @ecavallo

jonsterling: (this just replicates the behavior of `master`)

cangiuli: This is a great first step, but as we discussed, some way of defining an `all.red` file would avoid a lot of re-checking.

jonsterling: @cangiuli Yes; since defining this `all.red` file in a non-stupid way requires some new functionality, I suggest that we do that on a different branch and get this merged shortly.

library/README.md
```diff
@@ -14,3 +14,22 @@ We have no absolute rules, but try to adhere to the following:
 - Write multi-word semantic units with hyphen (`is-contr`, `weak-connection`)
 - Write subordinate semantic units with slash (`plus/assoc`, `symm/unit`)
 - Reserve `is-` and `has-` for h-propositions (`is-contr`, `has-hlevel`)
+
+Library structure
```
cangiuli: Thanks!

PR #404, opened 2018-10-16
Clean up the lexer and highlighter with Thoughts

favonia:


vim/syntax/redtt.vim
```diff
@@ -1,7 +1,9 @@
 " vim-redtt syntax
 " Language:     redtt
 " Author:       Carlo Angiuli
-" Last Change:  2018 October 15
+" Last Change:  2018 October 16
+"
+" Note:         slightly modified by Favonia
```
cangiuli: You can just add your name to the Author list if you want.

favonia: LOL I just want to make sure others know whom to blame.

PR #413, opened 2018-10-17
really remove modal stuff, fix exit codes

jonsterling:
@favonia why was Travis accepting the previous PRs??


src/frontend/Frontend.ml
```diff
@@ -94,7 +94,8 @@ let load options source =
   | ParseError.E (posl, posr) ->
     let loc = Some (posl, posr) in
     let pp fmt () = Format.fprintf fmt "Parse error" in
-    Log.pp_message ~loc ~lvl:`Error pp Format.err_formatter ()
+    Log.pp_message ~loc ~lvl:`Error pp Format.err_formatter ();
+    exit 1
```
jonsterling: I forgot this. BTW why do we exit with code `-1` elsewhere? What does that mean?

favonia: I regret my thoughtless random change. Please unify them to 1 or I'll do so.

jonsterling: done!

PR #414, opened 2018-10-17
Use names instead of strings to look up a datatype.

favonia:
@jonsterling I also change the name in ML from `string` to `mlval`.

### DIFF
--- a/src/frontend/Grammar.mly
+++ b/src/frontend/Grammar.mly
@@ -587,14 +587,24 @@ tm:
       Tm.Cons (e0 env, e1 env) }
 
   | LPR; DATA; dlbl = ATOM; RPR
-    { fun _ ->
+    { fun env ->
       make_node $startpos $endpos @@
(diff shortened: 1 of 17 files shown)


jonsterling:
Thanks! I'll review this shortly.

src/frontend/Grammar.mly
```diff
@@ -587,12 +589,12 @@ tm:
   | LPR; DATA; dlbl = ATOM; RPR
     { fun _ ->
       make_node $startpos $endpos @@
-      Tm.Data {lbl = dlbl; params = []} }
+      Tm.Data {lbl = Name.named (Some dlbl); params = []} }
```
jonsterling:
```suggestion
      begin
        match R.get dlbl env with
        | `Datatype alpha ->
          Tm.Data {lbl = alpha; params = []} }
        | _ -> failwith "Taste it!"
      end
```

PR #417, opened 2018-10-18
random cleanup

jonsterling:


src/frontend/Frontend.ml
```diff
@@ -52,20 +53,24 @@ struct
...
-        bind (Elab.eval_cmd mlcmd.con) @@ fun _ ->
-        bind (report_unsolved ~loc:mlcmd.span) @@ fun _ ->
-        bind resolver @@ fun res ->
-        bind get_per_process @@ fun per_process ->
-        ret (res, per_process)
+        Elab.eval_cmd mlcmd.con >>
+        report_unsolved ~loc:mlcmd.span >>
+        resolver >>= fun res ->
+        get_per_process <<@> fun per_process ->
+          res, per_process
```
favonia: Why do we want extra indentation in the last line?

favonia:
Actually, let me merge this PR now so that I can work on `Frontend`.

src/frontend/Frontend.ml
jonsterling: Personally, I don't want the extra indentation. But the indentation policy is to do whatever `ocp-indent` does.

PR #418, opened 2018-10-18
[Rot] Stage 1

favonia:
Bonus:
- [x] warning about the quick mode.


src/frontend/RotData.ml
```diff
@@ -0,0 +1,26 @@
+let version : string = "where do Thoughts come from?"
```
jonsterling: Where do correct ideas come from? Do they drop from the skies? No. Are they innate in the mind? No. They come from social practice, and from it alone; they come from three kinds of social practice, the struggle for production, the class struggle and scientific experiment. It is man’s social being that determines his thinking. Once the correct ideas characteristic of the advanced class are grasped by the masses, these ideas turn into a material force which changes society and changes the world. In their social practice, men engage in various kinds of struggle and gain rich experience, both from their successes and from their failures. Countless phenomena of the objective external world are reflected in a man’s brain through his five sense organs  —  the organs of sight, hearing, smell, taste and touch. At first, knowledge is perceptual. The leap to conceptual knowledge, i.e., to ideas, occurs when sufficient perceptual knowledge is accumulated. This is one process in cognition. It is the first stage in the whole process of cognition, the stage leading from objective matter to subjective consciousness from existence to ideas. Whether or not one’s consciousness or ideas (including theories, policies, plans or measures) do correctly reflect the laws of the objective external world is not yet proved at this stage, in which it is not yet possible to ascertain whether they are correct or not. Then comes the second stage in the process of cognition, the stage leading from consciousness back to matter, from ideas back to existence, in which the knowledge gained in the first stage is applied in social practice to ascertain whether the theories, policies, plans or measures meet with the anticipated success. Generally speaking, those that succeed are correct and those that fail are incorrect, and this is especially true of man’s struggle with nature. In social struggle, the forces representing the advanced class sometimes suffer defeat not because their ideas are incorrect ! but because, in the balance of forces engaged in struggle, they are not as powerful for the time being as the forces of reaction; they are therefore temporarily defeated, but they are bound to triumph sooner or later. Man’s knowledge makes another leap through the test of practice. This leap is more important than the previous one. For it is this leap alone that can prove the correctness or incorrectness of the first leap in cognition, i.e., of the ideas, theories, policies, plans or measures formulated in the course of reflecting the objective external world. There is no other way of testing truth. Furthermore, the one and only purpose of the proletariat in knowing the world is to change it. Often, correct knowledge can be arrived at only after many repetitions of the process leading from matter to consciousness and then back to matter, that is, leading from practice to knowledge and then back to practice. Such is the Marxist theory of knowledge, the dialectical materialist theory of knowledge. Among our comrades there are many who do not yet understand this theory of knowledge. When asked the sources of their ideas, opinions, policies, methods, plans and conclusions, eloquent speeches and long articles they consider the questions strange and cannot answer it. Nor do they comprehend that matter, can be transformed into consciousness and consciousness into matter, although such leaps are phenomena of everyday life. It is therefore necessary to educate our comrades in the dialectical materialist theory of knowledge, so that they can orientate their thinking correctly, become good at investigation and study and at summing up experience, overcome difficulties, commit fewer mistakes, do their work better, and struggle hard so as to build China into a great and powerful socialist country and help the broad masses of the oppressed and exploited throughout the world in fulfillment of our great internationalist duty.

Mao Zedong, 1963

src/frontend/Dev.ml
```diff
@@ -9,7 +9,12 @@ type twin = [`Only | `TwinL | `TwinR]
 
 type 'a decl =
   | Hole of [`Rigid | `Flex]
-  | Defn of ResEnv.visibility * [`Transparent | `Opaque] * 'a
+  | Macro of 'a
```
jonsterling: Based on our chat, I see what this does and it looks like a good idea! Maybe another name would help, like `Auxiliary`.

src/frontend/Dev.mli
```diff
@@ -6,8 +6,19 @@ type ty = Tm.tm
...
+    (** this means a variable can be expanded into a term,
+        and this is not generated by the user. *)
+  | UserDefn of
+    {source : FileRes.filepath;
+     visibility : ResEnv.visibility;
+     opacity : [`Transparent | `Opaque];
+     tm : 'a}
+    (** this is a definition given by the user and has been type-checked *)
   | Guess of {ty : 'a; tm : 'a}
+    (** this means the Unify is guessing some term [tm] of type [ty],
```
jonsterling: let's change this comment to say something like:

> we have a term [tm] of type [ty], which is not yet the same as the hole we are trying to fill

PR #420, opened 2018-10-19
simplify univalence proof

ecavallo:
down with connections!!

### DIFF
--- a/library/prelude/univalence.red
+++ b/library/prelude/univalence.red
@@ -9,41 +9,41 @@ def ua (A B : type) (e : equiv A B) : path^1 type A B =
 def univalence (B : type) : is-contr^1 ((A : type) × equiv A B) =
   ( (B, id-equiv B)
   , λ (A,e) i →
-      let VB : type = V i A B e in
-      let proj/B (g : VB) : B = g .vproj in
-      ( _
-      , proj/B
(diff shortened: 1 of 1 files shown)


jonsterling:
cool!

ecavallo:
The indentation under `, λ` was/is not consistent in here, but I forget which one is JST.

jonsterling:
> The indentation under `, λ` was/is not consistent in here, but I forget which one is JST.

JST is to not indent underneath lambda, or anything that binds everytihng to its right.

jonsterling:
I was fiddling with this file and I'm noticing a LOT of mortality in either the elaborator or the typechecker concerning `V` types. I hope at least some of it is resolved by my new type checker.

library/prelude/univalence.red
```diff
@@ -3,41 +3,47 @@ import prelude.connection
 import prelude.hlevel
 import prelude.equivalence
 
-def ua (A B : type) (E : equiv A B) : path^1 type A B =
-  λ i →
-    `(V i A B E)
+def ua (A B : type) (e : equiv A B) : path^1 type A B =
+  λ i → V i A B e
```
favonia: I thought I already updated this LOL!

PR #424, opened 2018-10-20
tiny tweak to fix a grammar error in error message

jozefg:
I know it's trivial but I noticed it when messing around.


jonsterling:
thanks!

PR #446, opened 2018-10-25
added WIP GCD stuff

kaonn:


library/cool/gcd.red
```diff
@@ -0,0 +1,181 @@
...
+          elim (sub m (suc n)) [
+          | zero → zero
+          | suc _ → suc n
+          ]
+        | suc _ → (f (sub (suc n) (suc m')) (sub/le n m')) (suc m')
+        ]
+      ]
+    )
+
+def const/zero : nat → nat =
```
ecavallo: this is a pretty confusing name!

kaonn: what about always/zero?

ecavallo: isn't this the identity function, though?

kaonn: oops you're right. eta-expansion?

ecavallo: maybe id/nat? the next definition is already named eta...

kaonn: I was gonna change it to eta-equiv, but that works (commuting conversion is the right word probably)

jonsterling:
Shall we merge it?

PR #447, opened 2018-10-26
add bool paths, DNE, LEM, aux proofs, refactor

clayrat:


jonsterling:
@clayrat Wow, this is super cool! At redtt HQ, we're all super impressed with how you pulled this together :sunglasses:

library/cool/logic.red
```diff
@@ -0,0 +1,36 @@
...
+import basics.hedberg
+
+def no-double-neg-elim (f : (A : type) → stable A) : void = 
+  let f2 = f bool in 
+
+  -- transport along the path induced from `not` by univalence
+  let tf2 = coe 0 1 f2 in λ i → stable (not/path i) in
+
+  -- transporting a dependent function produces a path to the original
+  let apdf : path _ tf2 f2 = apd^1 _ stable f _ _ not/path in
```
jonsterling: Can we inline the use of `apd` here?

library/paths/bool.red
```diff
@@ -0,0 +1,38 @@
...
+    | ff → void
+    ]
+  | ff → elim [
+    | tt → void
+    | ff → unit
+    ]
+  ]
+
+def bool-refl : (x : bool) → bool-path/code x x =
+  elim [
```
jonsterling: I think this can be made even shorter by writing `\ * -> triv`; the idea is that `*` is a special "inversion pattern" which calls the eliminator, and then provides the rhs to each branch.

library/paths/bool.red
```diff
@@ -0,0 +1,38 @@
...
+  | tt → triv
+  | ff → triv
+  ]  
+
+def bool-path/encode (x y : bool) (p : path bool x y) : bool-path/code x y =
+  coe 0 1 (bool-refl x) in λ i → bool-path/code x (p i)  
+
+def not/neg : (x : bool) → neg (path bool (not x) x) = 
+  elim [ 
+   | tt → bool-path/encode ff tt
```
jonsterling: The convention in code like this is to align the pipe with the `e` in `elim`.

library/prelude/path.red
```diff
@@ -34,6 +34,20 @@ def funext
   λ i x →
   p _ i
 
+def happly 
+  (A : type) 
+  (B : A → type) 
+  (f g : (x : A) → B x) 
+  (p : path ((x : A) → B x) f g) 
+  : (x : A) → path (B x) (f x) (g x) = 
+  λ x i → p i x
```
jonsterling: In Cubical Thought, something like this should just be inlined; then we don't need to pass all these extra arguments to `happly` just to get a function application.

library/prelude/path.red
```diff
@@ -34,6 +34,20 @@ def funext
...
+  (f g : (x : A) → B x) 
+  (p : path ((x : A) → B x) f g) 
+  : (x : A) → path (B x) (f x) (g x) = 
+  λ x i → p i x  
+
+def apd 
+  (A : type) (P : A → type) 
+  (f : (x : A) → P x) (x y : A) (p : path A x y) 
+  : path (P y) (coe 0 1 (f x) in λ i → P (p i)) (f y) = 
+  λ i → coe i 1 (f (p i)) in λ j → P (p j)
```
jonsterling: If we inline the use of `apd`, we can delete this definition.

library/prelude/path.red
```diff
@@ -105,6 +119,18 @@ def symmd (A : 𝕀 → type) (p : (i : 𝕀) → A i) : pathd (symm^1 _ A) (p 1
   | i=1 → refl
   ]
 
+-- transporting backwards is transporting forwards along inverted path (up to composition)
+def coe/symm/d (A : type) (P : A → type) (x y : A) (p : path A x y) (py : P y) :
```
jonsterling: nicely done :sunglasses:

library/prelude/path.red
clayrat: I think I'm not actually using this anywhere, just added it for symmetry with `funext`.

library/prelude/path.red
clayrat: Sure, just thought to include it since it features in the book regularly later.

library/prelude/path.red
clayrat: yeah I spent some time figuring this one out 😓

clayrat:
Made the changes, so should I also remove `happly` and `apd`? I must say the inlined version with `apd` does look more verbose, since you have to at least mention the path twice.

clayrat:
@jonsterling removed

PR #453, opened 2018-10-28
rename triv to ★, random cleanup

jonsterling:


library/paths/bool.red
```diff
@@ -6,27 +6,26 @@ import basics.isotoequiv
...
-  coe 0 1 (bool-refl x) in λ i → bool-path/code x (p i)  
+  coe 0 1 (bool-refl x) in λ i → bool-path/code x (p i)
 
-def not/neg : (x : bool) → neg (path bool (not x) x) = 
-  elim [ 
-  | tt → bool-path/encode ff tt 
-  | ff → bool-path/encode tt ff 
-  ]
+def not/neg : (x : bool) → neg (path bool (not x) x) =
+  λ * → bool-path/encode _ _
```
jonsterling: This is ^^cool^^

library/cool/moebius-boundary.red
```diff
@@ -79,24 +81,14 @@ def s1→moebius-boundary→s1 :
   | loop i → λ j → s1→moebius-boundary→s1/loop i j
   ]
 
-quit
+def test (i : 𝕀) : moebius-boundary =
+  s1→moebius-boundary (loop i)
 
--- there is an invalid fhcom in the middle?!
```
jonsterling: I checked and this no longer occurs; I guess we fixed that bug somehow.

favonia: I fixed the bug but the new updates on this file are stacked up in another PR (which was then blocked by another bug).

jonsterling: @favonia Ah, ok.

PR #422, opened 2018-10-19
^cooler^ invariance example

cangiuli:
With lots of help from @jonsterling and @ecavallo.

PR #456, opened 2018-10-29
Unary and binary natural numbers.

cangiuli:
I'm moving this out of #422. The representation is wrong, because `nil` and `cons0 nil` are both zero. Anyone (@ecavallo ?) is welcome to hack on this, but if nobody else does I'll eventually get back to it.

### DIFF
--- a/library/cool/nats.red
+++ b/library/cool/nats.red
@@ -1,0 +1,53 @@
+import prelude
+import data.nat
+import basics.isotoequiv
+
+/-
+data nat where
+| zero
+| suc (x : nat)
+-/
+
+data binnat where
+| nil
+| cons0 (x : binnat)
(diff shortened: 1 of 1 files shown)


clayrat:
In Coq and https://github.com/sbp/idris-bi they use 1 as a base constructor, and then construct nats and ints separately on top of that.

library/cool/nats.red
```diff
@@ -0,0 +1,77 @@
+import prelude
+import data.nat
+import basics.isotoequiv
+
+/-
```
jonsterling: let's remove this comment

cangiuli:
OMG!! I should open PRs with broken code more often! Thank you @favonia!

PR #458, opened 2018-11-01
Detect cyclic module dependency.

favonia:


cangiuli:
Can we generalize this to arbitrary hlevel?

PR #461, opened 2018-11-06
print mismatched names in Quote

clayrat:
Helps when debugging


src/core/Quote.ml
```diff
@@ -2,6 +2,7 @@ open Domain
 open RedBasis
 open Bwd
 open BwdNotation
+open Name
```
jonsterling: Let's not open this, and instead call `Name.to_string`

PR #462, opened 2018-11-12
[vim] Usability improvements.

cangiuli:
This contains three improvements to the vim mode.

- `<LocalLeader>L` and `<LocalLeader>P` add the `--ignore-cache` option.
- Typing `<LocalLeader>l` or `<LocalLeader>L` in the `redtt` output buffer will recheck the most recently checked `.red` file, if it is still open. (Requested by @jonsterling)
- When closing a `.red` buffer, only close the `redtt` output buffer if the `.red` file is the most recently checked one.

### DIFF
--- a/src/frontend/Importer.ml
+++ b/src/frontend/Importer.ml
@@ -57,10 +57,10 @@ struct
       | Some rot -> ret rot
       | None ->
         begin
-          if !ignore_rot then ret None else begin
-            let rotpath = FileRes.stem_to_rot stem in
+          if !ignore_rot then
+            ret None
+          else
             RotIO.try_read ~redsum:None ~importer:import ~stem
-          end
         end >>= function
         | Some rot ->
           store_module stem rot >> ret rot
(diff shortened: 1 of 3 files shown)


favonia:
I have a separate request: adding a new binder to ignore .rot files ("--ignore-cache"). Maybe `<LocalLeader>L`?

cangiuli:
@favonia Done. Right after pushing I realized...do you also want `<LocalLeader>P` for checking a partial file with `--ignore-cache`? (It's trivial to add.)

favonia:
There is no reason for me to say no! LOL

By the way, once redtt becomes mature, maybe we do not need these special keybindings anymore.

jonsterling:
I'm going to try and test this today; I have only several free minutes though, so we'll see.

jonsterling:
@cangiuli Is it intentional that when I run `\l` from inside the output buffer, it moves my cursor to the redtt buffer?

cangiuli:
Yes, but I could be convinced to implement some other behavior. The problem with staying in the output buffer is that I'm not sure where to place the cursor, given that the content is changing completely. The top? The same line number as before? Neither seems satisfactory.

jonsterling:
@cangiuli That's a good point! I'm ok with this behavior. I found it a little surprising, but it doesn't bother me.

PR #464, opened 2018-11-13
some facts about smash product

ecavallo:
Don't think I have anything feasible left to do on this branch.

- prove that smash is commutative
- prove that bool is a unit for smash
- define the associator (but don't prove it's an equivalence)

I moved `cool.pointed` => `basics.pointed` also.


favonia:
Essentially you are growing the core library. :smiley_cat:

library/pointed/smash.red
```diff
@@ -0,0 +1,264 @@
...
+
+-- absorption
+
+def smash-unit/contr (X : ptype) : is-contr (smash X punit) =
+  let contr/proj (a : X .fst) : (u : unit)
+    → path (smash X punit) (proj a u) (proj (X .snd) ★)
+    =
+    elim [★ → pivotr X punit a (X .snd)]
+  in
+  let contr/gluel : (u : unit) → [i j] smash X punit [
```
cangiuli: omg, a punit square!

PR #465, opened 2018-11-15
Misc lemmas

clayrat:
Since it looks like I'm stuck in #457 without a box tactic, here are some helper lemmas I've written along the way.

### DIFF
--- a/library/basics/hedberg.red
+++ b/library/basics/hedberg.red
@@ -48,7 +48,7 @@ def paths-stable→set (A : type) (st : (x y : A) → stable (path A x y)) : is-
 def discrete→set (A : type) (d : discrete A) : is-set A =
   paths-stable→set A (λ x y → dec→stable (path A x y) (d x y))
 
-def mere-relation/set-equiv 
+def hrel/set-equiv 
   (A : type) (R : A → A → type) 
   (R/prop : (x y : A) → is-prop (R x y))
   (R/refl : (x : A) → R x x) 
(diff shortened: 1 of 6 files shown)


ecavallo:
For now, it is possible to indirectly write a `box` with a composition:
```
def boxlike
  (A : 𝕀 → type) (B0 : [j] type [j=0 → A 0]) (B1 : [j] type [j=0 → A 1])
  (a : (i : 𝕀) → A i) (b0 : [j] B0 j [j=0 → a 0]) (b1 : [j] B1 j [j=0 → a 1])
  (i : 𝕀)
  : comp 0 1 (A i) [i=0 → B0 | i=1 → B1]
  =
  comp 0 1 (a i) in λ j → comp 0 j (A i) [i=0 → B0 | i=1 → B1] [
  | i=0 → b0
  | i=1 → b1
  ]
```

library/basics/hedberg.red
```diff
@@ -46,3 +47,27 @@ def paths-stable→set (A : type) (st : (x y : A) → stable (path A x y)) : is-
 -- Hedberg's theorem for decidable path types
 def discrete→set (A : type) (d : discrete A) : is-set A =
   paths-stable→set A (λ x y → dec→stable (path A x y) (d x y))
+
+def mere-relation/set-equiv
```
jonsterling: @mortberg I'm not a fan of the terminology `mere-relation`; what is preferred in your Thought?

mortberg: I never liked the word "mere" for hProp valued things... Why not just call it `relation` or `rel`?

clayrat: For me simply "relation" is anything of type `A → B → type`. In [quotient.red](https://github.com/RedPRL/redtt/blob/master/library/cool/quotient.red#L61) a propositional relation is called `prop-valued`, is that good enough?

mortberg: Hmm, what about `hrel` or `hrelation`? (just like we use `hProp` for the universe of h-propositions)

clayrat: @jonsterling any preference? :)

mortberg: `hrel` is the name that Voevodsky used: https://github.com/UniMath/UniMath/blob/master/UniMath/Foundations/Sets.v#L474

I would go for that

jonsterling: @mortberg Sounds good to me!

PR #470, opened 2018-11-24
use cap tactic in paths.s2

ecavallo:
It's not a performance win (`s2/decode` actually takes much more time to check now) but the normal forms are shorter.


ecavallo:
This incidentally made it easier to define the `oloop` line of equivalences directly instead of using `is-equiv/prop`, which improves checking time considerably.

PR #474, opened 2018-12-06
successor on biinv-int which cancels pred exactly

ecavallo:
@mortberg 

Turns out it is entirely possible, just requires some IH-strengthening. I did `suc`, but `pred` wouldn't be any harder.


mortberg:
Woah! And you convinced me that it wasn't possible today...

mortberg:
Can you use these integers in the proof that Omega(S1) = Int ? If I remember correctly I think it will simplify the loop case of https://github.com/RedPRL/redtt/blob/master/library/paths/s1.red#L106

mortberg:
I think my idea was that you don't need to compose with pred-isuc as it's definitional

ecavallo:
Well, you would have to rewrite everything that does case analysis on `int` first...

mortberg:
Yeah, my prediction is that that will be quite a nightmare with these integers...

PR #482, opened 2019-06-08
Parametric smash

ecavallo:
A characterization of polymorphic maps X ∧ Y →* X ∧ Y, assuming a certain axiom that would follow from internal parametricity.

### DIFF
--- a/library/cool/parametric-smash.red
+++ b/library/cool/parametric-smash.red
@@ -1,0 +1,164 @@
+import prelude
+import prelude.pointed
+import data.bool
+import data.smash
+import basics.retract
+import paths.bool
+import pointed.bool
+import pointed.smash
+
+-- characterization of the maps of type (X Y : ptype) → X ∧ Y →* X ∧ Y,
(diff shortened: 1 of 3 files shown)


library/cool/parametric-smash.red
```diff
@@ -0,0 +1,164 @@
...
+def pconst (X Y : ptype) : pmap X Y =
+  (λ _ → Y .snd , λ _ → Y .snd)
+
+def has-square-fillers (A : type) : type =
+  (p0- p1- : 𝕀 → A)
+  (p-0 : path A (p0- 0) (p1- 0))
+  (p-1 : path A (p0- 1) (p1- 1))
+  → [i j] A [i=0 → p0- j | i=1 → p1- j | j=0 → p-0 i | j=1 → p-1 i ]
+
+def set→has-square-fillers (A : type) (A/set : is-set A)
```
jonsterling: I wonder if such lemmas should be moved elsewhere? I don't care too much though. BTW This looks like a useful theorem :)

Issue #486, opened 2020-01-07
Upgrade compiler and dependencies


jonsterling:
- [ ] OCaml version
- [ ] Latest Dune version
- [ ] Upgrade all dependencies (to be found in `redtt.opam`)

PR #487, opened 2020-01-09
updating versions

ivoysey:
this ought to address #486!

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ end up using one of these for the lexer at one point.
 
 | prerequisite |      | version                                                                | how to install                  |
 | ------------ | ---- | :--------------------------------------------------------------------- | ------------------------------- |
-| OPAM         | `>=` | [`2.0.1`](https://github.com/ocaml/opam/releases/tag/2.0.1)            | manually or via package manager |
+| OPAM         | `>=` | [`2.0.5`](https://github.com/ocaml/opam/releases/tag/2.0.5)            | manually or via package manager |
 
 If this is your first time configuring OPAM, please run `opam init`; this will
 automatically install OCaml. If you are using a version of OCaml other than
--- a/README.md
+++ b/README.md
@@ -69,11 +69,11 @@ end up using one of these for the lexer at one point.
 
 | prerequisite |      | version                                                                | how to install                  |
 | ------------ | ---- | :--------------------------------------------------------------------- | ------------------------------- |
-| Opam         | `>=` | [`2.0.1`](https://github.com/ocaml/opam/releases/tag/2.0.1)            | manually or via package manager |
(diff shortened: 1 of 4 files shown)


jonsterling:
I wonder if this newfangled "GitHub Actions" stuff can provide a more reliable alternative to travis ??

jonsterling:
https://github.com/avsm/hello-world-action-ocaml/blob/master/.github/workflows/test.yml looks kind of cool.

ivoysey:
oh yeah. maybe. i know only enough about travis to get in trouble, but it's simple enough yaml pushing. if there's something better that's fine too.

jonsterling:
@ivoysey Yeah lol, when I set up the travis I basically was like this dog: 

image

ivoysey:
so it looks like the build correctly picked up the new env vars i set to increment the versions of things but failed on a wget command. from the bottom of the log:

```
--2020-01-09 21:42:46--  https://github.com/ocaml/opam/releases/download/2.1.1/opam-2.1.1-x86_64-linux

Resolving github.com (github.com)... 140.82.114.3

Connecting to github.com (github.com)|140.82.114.3|:443... connected.

HTTP request sent, awaiting response... 404 Not Found

2020-01-09 21:42:46 ERROR 404: Not Found.

The command "sh <(curl -sL https://raw.githubusercontent.com/ocaml/ocaml-ci-scripts/master/.travis-ocaml.sh)" failed and exited with 8 during .
```

cangiuli:
Well, the latest release of opam is apparently 2.0.5, so that would explain why the wget failed.

ivoysey:
ah, good catch. i must have had the wrong thing in my paste buffer. just fixed that, and this maybe triggered another travis build.

cangiuli:
By the way, although the build passed, I noticed that the output of `opam pin add -y redtt .` complains that the `opam-version` in redtt.opam is wrong.

ivoysey:
yeah, it looks like it doesn't want you to specify minor versions.

ivoysey:
ok, i think that's got it squared away. it also does seem like it goes to the latest minor revision of the major revision you specify, fwiw, so it still got built on 2.0.5 but you just specify less than that.

README.md
```diff
@@ -69,11 +69,11 @@ end up using one of these for the lexer at one point.
 
 | prerequisite |      | version                                                                | how to install                  |
 | ------------ | ---- | :--------------------------------------------------------------------- | ------------------------------- |
-| Opam         | `>=` | [`2.0.1`](https://github.com/ocaml/opam/releases/tag/2.0.1)            | manually or via package manager |
+| OPAM         | `>=` | [`2.0.1`](https://github.com/ocaml/opam/releases/tag/2.0.1)            | manually or via package manager |
```
jonsterling: should the version documented here be changed?

ivoysey: it certainly should!

PR #489, opened 2021-11-27
Enable type inference for Menhir.

fpottier:
This is already useful now, and will be more useful in the future (Menhir's upcoming new code back-end will require `--infer`).


favonia:
Hmm Travis is no longer working, it seems.

favonia (approved):
LGTM

